Issue #64, opened 2024-07-23
replace the `R` with the paragraph symbol


yegor256:
In `policy.scss`, let's replace `content: "R"` with `content: "§"`. I tried to do this, but it didn't work. For some reason, the counter stops working and all rules are numbered as `§1`.

PR #170, opened 2024-11-30
replace the R with the paragraph symbol

Legolass322:
About https://github.com/zerocracy/pages-action/issues/64

I checked safari and chromium based browser is ok. Probably, @yegor256 you previously checked under some specific conditions. I assume that it might be not the case if you check it again. Otherwise, if we will define the configuration when the problem occurs, it will be easier to solve.

### DIFF
--- a/sass/bylaws.scss
+++ b/sass/bylaws.scss
@@ -35,7 +35,7 @@ $em: #1c448e;
             break-inside: avoid;
 
             &::before {
-                content: "R" counter(par) ": ";
+                content: "§" counter(par) ": ";
                 counter-increment: par;
             }
         }


yegor256:
@Legolass322 thanks!

0crat:
@Legolass322 Thanks for the contribution! You've earned +4 points for this: +4 as a basis; -16 for the lack of code review; -16 for absolutely no comments posted by reviewers; -8 for too few (2) hits-of-code; +40 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running balance is [+4](https://zerocracy.github.io/judges-action/zerocracy-vitals.html).

Issue #171, opened 2024-12-01
`@import` in SASS is deprecated


yegor256:
I see this, when running `make`:

```
+ sass --no-source-map --style=compressed --no-quiet --stop-on-error sass/main.scss target/css/main.css
Deprecation Warning: Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.

More info and automated migrator: https://sass-lang.com/d/import

   ╷
25 │ @import 'awards.scss';
   │         ^^^^^^^^^^^^^
   ╵
    sass/main.scss 25:9  root stylesheet
```

Let's get rid of it.

PR #188, opened 2024-12-26
Replace sass @import rule with @use

katozaaaa:
#171 Good day! Fixed the rule for importing sass, there are no warnings now


yegor256:
@katozaaaa thanks!

Issue #72, opened 2024-07-28
check the validity of HTML after its generation


yegor256:
Let's add an extra validation step to `Makefile`, which will check the validity of generated HTML. Maybe we can use this tool: https://www.html-tidy.org/

yegor256:
@codeninja-ru yep

PR #215, opened 2025-03-12
check the validity of HTML after its generation

Legolass322:
Hello, @yegor256 

This pr based on #72 issue.

Couple of comments:
- I am using mac, so I am checking OS, before installing tidy
- For linux it's checking if it is in dockerenv `([ -f /proc/self/cgroup ] && grep -q ":" /proc/self/cgroup)`.
Typically, it's enough to check `-f /.dockerenv` and search for "docker" in "/proc/self/cgroup", but I am using colima as docker engine. So, I modified it a bit as methods above does not detect docker env.
- Finally, tidy can return 1, 2 as exit codes. 1 for warnings, which I've decided to ignore.

My checks are green. Could you check it?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@
 .SHELLFLAGS := -x -e -o pipefail -c
 SHELL := bash
 
+OS_NAME := $(shell uname -s | tr A-Z a-z)
+
 YAMLS = $(wildcard tests/*.yml)
 FBS = $(subst tests/,target/fb/,${YAMLS:.yml=.fb})
 HTMLS = $(subst fb/,html/,${FBS:.fb=.html})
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -49,6 +51,20 @@ target/html/%.html: target/output/%
 	while IFS= read -r xpath; do
 		xmllint --xpath "$${xpath}" "$$(dirname "$@")/$${n}-vitals.html" > /dev/null
 	done <<< "$${xpaths}"
+	set +e
```
yegor256, line 54: @Legolass322 this line is redundant, because we have `.SHELLFLAGS` already

Makefile
```diff
@@ -49,6 +51,20 @@ target/html/%.html: target/output/%
...
+		echo "$$(dirname "$@")/$${n}.html has errors"
+		exit 1
+	fi
+	tidy -e "$$(dirname "$@")/$${n}-vitals.html"
+	result=$?
+	if [ "$${result}" -eq "2" ]; then
+		echo "$$(dirname "$@")/$${n}-vitals.html has errors"
+		exit 1
+	fi
+	set -e
```
yegor256, line 67: @Legolass322 this one too

Makefile
Legolass322: @yegor256 It's not. I am turning off (`set +e`) one of the flags (and returning it back below `set -e`) while checking with tidy, because tidy returns 1 on tests, which drops the pipeline. Why 1? There are some minor warnings in tests. 

I've seen:
- "Warning: missing </span> before </td>" and "Warning: trimming empty <span>" in line `<td><span></td>`
- "Warning: <a> attribute "href" lacks value" in `<a href="" onclick="$('.p_yegor256').toggle(); return false;">2 awards</a>`

In my opinion, it's ok. So I am checking exit code manually on errors `if [ "$${result}" -eq "2" ];`
Is it good?

Legolass322: @yegor256 Works fine, thank you for suggestion

Legolass322:
@yegor256 Iteration 2 in second commit. Fixed comments above

yegor256:
@Legolass322 looks good, but there are merge conflicts now

Legolass322:
@yegor256 Fixed. Let's merge it first. Then I'll fix stylelint pr, and then html-minifier pr

yegor256:
@Legolass322 thanks!

PR #216, opened 2025-03-12
scss files are not checked by a linter

Legolass322:
Hello, @yegor256 

I've added stylelint with standart config. Except 2 rules:
- `at-rule-no-unknown` to avoid `@use` problems
- `media-feature-range-notation` transforms `@media(max-width: 1280px)` into `@media(width <= 1280px)`, which might has problems with current sass compiler

Could you check it?


Makefile
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: MIT
 
 .ONESHELL:
-.PHONY: clean all assets install rake
+.PHONY: clean all assets install rake fix-styles
```
yegor256, line 5: @Legolass322 maybe it's better to call it `stylelint`?

Legolass322:
@yegor256 Fixed comment above

Makefile
```diff
@@ -79,6 +82,7 @@ install: $(SAXON) | target
 	bundle install
 	npm --no-color install -g uglify-js
 	npm --no-color install -g sass@1.77.2
+	npm --no-color install -g stylelint stylelint-config-standard stylelint-scss
```
yegor256, line 85: @Legolass322 how about adding fixed versions for them all?

Legolass322:
@yegor256 Added versions to pkgs, including uglifyjs. I assume, it's ready to merge

yegor256:
@Legolass322 thanks!

PR #217, opened 2025-03-12
compress/minify HTML in entry.sh

Legolass322:
Hello, @yegor256 

I've added html-minifier. After a brief look I haven't understand why you wanted it only in `entry.sh ` with only vitals file generated (maybe misunderstanding). Anyways, here I've decided minify both html files.

Also, as you can see, in `xmllint` line I've changed the file it is linting. Somewhy xmllint producing parse errors, when html in single line, even so html is correct.

Is it ok? Should I only change vitals file?

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -79,6 +79,7 @@ fi
 
 # Build a summary HTML.
 html=${INPUT_OUTPUT}/${name}-vitals.html
+htmlmin=${INPUT_OUTPUT}/${name}-vitals.min.html
 java -jar "${SELF}/target/saxon.jar" \
     "-s:${INPUT_OUTPUT}/${name}.rich.xml" \
     "-xsl:${SELF}/target/xsl/vitals.xsl" \
@@ -90,5 +91,6 @@ java -jar "${SELF}/target/saxon.jar" \
     "logo=${INPUT_LOGO}" \
     "css=$(cat "${SELF}/target/css/main.css")" \
     "js=$(cat "${SELF}/target/js/main.js")"
-echo "HTML generated at ${html}"
+html-minifier "${html}" --config-file "${SELF}/html-minifier-config.json" -o "${htmlmin}"
+echo "HTML generated at ${html}, minified version at ${htmlmin}"
 rm "${INPUT_OUTPUT}/${name}.rich.xml"
--- a/entry.sh
+++ b/entry.sh
@@ -79,7 +79,6 @@ fi
 
 # Build a summary HTML.
 html=${INPUT_OUTPUT}/${name}-vitals.html
-htmlmin=${INPUT_OUTPUT}/${name}-vitals.min.html
 java -jar "${SELF}/target/saxon.jar" \
     "-s:${INPUT_OUTPUT}/${name}.rich.xml" \
     "-xsl:${SELF}/target/xsl/vitals.xsl" \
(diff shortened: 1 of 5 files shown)


yegor256:
@Legolass322 the minification you added doesn't make sense, since it's only visible during testing (when we run `make`). Instead, we should minify the file generated in `entry.sh` by this line: https://github.com/zerocracy/pages-action/blob/master/entry.sh#L82 An HTML is created, but it's not minified. We should minify it.

Legolass322:
@yegor256 Yes, thank you. I misunderstood the pipeline. Will move it to entry.sh

entry.sh
```diff
@@ -90,5 +91,6 @@ java -jar "${SELF}/target/saxon.jar" \
     "logo=${INPUT_LOGO}" \
     "css=$(cat "${SELF}/target/css/main.css")" \
     "js=$(cat "${SELF}/target/js/main.js")"
-echo "HTML generated at ${html}"
+html-minifier "${html}" --config-file "${SELF}/html-minifier-config.json" -o "${htmlmin}"
```
yegor256, line 94: @Legolass322 no need to keep the original HTML document. Let's just minify it in place.

Legolass322: @yegor256 About that. I was thinking how to deal with xmllint. Xml parser is more strict that an html's. Correct html5 document produces parse errors [here](https://github.com/zerocracy/pages-action/blob/master/Makefile#L50), because minifier makes it not xml-readable (removes some backslashes).

Legolass322: I can instead of producing *.html and *.min.html, produce *.html (minified) and *-unminified.html. And check the last one with xmllint.

yegor256: @Legolass322 maybe you can minify and preserve XML validity?

Legolass322: @yegor256 It looks like there is an option I missed. Html-minifier has `keepClosingSlash` configuration option, which probably is enough to produce xml valid documents. At least it makes with tests.

Legolass322:
@yegor256 Found this in logs
```
npm WARN EBADENGINE   package: 'eslint@9.22.0',
npm WARN EBADENGINE   required: { node: '^18.18.0 || ^20.9.0 || >=21.1.0' },
npm WARN EBADENGINE   current: { node: 'v12.22.12', npm: '7.5.2' }
npm WARN EBADENGINE }
```

Does rultor has such an old version?)

I guess sass version was selected just to be runnable too
```
+ npm --no-color install -g sass@1.77.2
K?25hnpm.WARN.EBADENGINEdUnsupportedrenginea{ge:lsillcplaceDepenode_modules/sas0mK
npm WARN EBADENGINE   package: 'sass@1.77.2',
npm WARN EBADENGINE   required: { node: '>=14.0.0' },
npm WARN EBADENGINE   current: { node: 'v12.22.12', npm: '7.5.2' }
npm WARN EBADENGINE }
K?25h##############].\.reify:sass:ahttphfetchfGETh200Thttps://registry.npmjs.or0mKm0mK
added 18 packages, and audited 19 packages in 1s
```

Legolass322:
@yegor256 As I can see, it didn't cause problems last time due to you merged it manually (eslint pr).
Is it possible to update node version in rultor to have opportunity use it with this project?

yegor256:
@Legolass322 this is how we install nodejs: https://github.com/zerocracy/pages-action/blob/master/.rultor.yml#L10 what do you suggest changing?

Legolass322:
@yegor256 I see couple of options here:
- Use same way as in Dockerfile (only if rultor is running on os which satisfies [this](https://github.com/nodesource/distributions?tab=readme-ov-file#deb-supported-versions))
```sh
curl -fsSL https://deb.nodesource.com/setup_18.x -o nodesource_setup.sh
sudo -E bash nodesource_setup.sh
sudo apt-get install -y nodejs
```
- Another option for unix based is https://github.com/nvm-sh/nvm
```sh
curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.2/install.sh | bash
# or
# wget -qO- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.2/install.sh | bash
nvm install 18
nvm use 18
```
With nvm you could alse add .nvmrc file which will contain concrete node version and use it in pipelines, docker, etc

yegor256:
@rultor merge

Legolass322:
@yegor256 I've added node 18 setup into rultor.yaml. Please, check it before call.

yegor256:
@Legolass322 thanks!

Issue #82, opened 2024-07-30
check `.js` files with eslint


yegor256:
Now, `.js` files are not validated by any static analyzer. Let's add eslint to the build.

codeninja-ru:
@yegor256 Should this be closed?

yegor256:
@codeninja-ru yep

PR #218, opened 2025-03-12
check .js files with eslint

Legolass322:
Hello, @yegor256 

This pr related to #82 

I've added eslint. And also changed node version from 16 to 18 in make workflow, due to these reasons:
- it correlates with docker, which is installing 18th version
- eslint with node 16 does not support lots of rules, such as "semi"

What about rules, standart configs? What do you prefer?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ $(SAXON): | target
 
 install: $(SAXON) | target
 	bundle install
-	npm --no-color install -g eslint
+	npm --no-color install -g eslint@9.22.0
 	npm --no-color install -g uglify-js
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -77,6 +78,7 @@ $(SAXON): | target
 
 install: $(SAXON) | target
 	bundle install
+	npm --no-color install -g eslint
```
yegor256, line 81: @Legolass322 maybe it's good to specify eslint version here?

Legolass322: Yes, let me add current version 9.2.2

yegor256:
@Legolass322 thanks!

PR #387, opened 2025-09-30
Add css validation

kodsurfer:
## Summary by CodeRabbit

* **Chores**
  * Added automated CSS validation in CI that runs on pushes and pull requests to catch styling issues early.
  * Introduced a local stylelint build step to ensure dependencies are installed and run linting with automatic fixes for stylesheets.
  * No direct UI changes; improves style consistency and reduces visual regressions.


kodsurfer:
@yegor256 work in progress, I hope to finish soon

Makefile
```diff
@@ -67,8 +67,11 @@ target/fb/%.fb: tests/%.yml Makefile | target/fb
 rake: $(SAXON) $(HTMLS)
 	bundle exec rake
 
-stylelint: sass/*.scss
-	stylelint sass/*.scss --fix
+node_modules/.bin/stylelint:
+	npm install
```
kreinba: `npm install` with no arguments needs a `package.json` in the repo root, and this repo does not ship one — `makes/install.sh` installs stylelint globally (`npm install -g stylelint@16.15.0 ...`). The `make` and `validate-css` jobs on this PR both fail here with `npm error code ENOENT ... package.json`. Drop this local install rule and keep invoking the global `stylelint`, or ship a `package.json` declaring the linter as a dev dependency.

Makefile
```diff
@@ -67,8 +67,11 @@ target/fb/%.fb: tests/%.yml Makefile | target/fb
...
+
+stylelint: node_modules/.bin/stylelint
+	./node_modules/.bin/stylelint stylelint sass/*.scss --fix
```
kreinba: The new command passes a literal `stylelint` argument before the SCSS glob, so the linter treats `stylelint` as a target path and fails before it touches any `.scss` file. The previous line was `stylelint sass/*.scss --fix`.

```suggestion
	./node_modules/.bin/stylelint sass/*.scss --fix
```

.github/workflows/csscheck.yml
```diff
@@ -0,0 +1,21 @@
...
+on:
+  push:
+  pull_request:
+
+jobs:
+  validate-css:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build CSS
```
kreinba: This job runs `make assets` directly after `actions/checkout`, with no setup of Ruby, `bundle`, or the global npm tools (`sass`, `eslint`, `uglify-js`, `stylelint`) that the `assets` target shells out to. The `make.yml` workflow in the same directory calls `./makes/install.sh` after setting up Ruby and Node — mirror that pattern here, otherwise the job cannot reach the validator.

kreinba (changes requested):
The new local stylelint install rule breaks the build for every CI job that runs `make`: `npm install` has no `package.json` to read, so the `make` and `validate-css` checks on this PR both die at `Makefile:71` with `npm error code ENOENT`. The new validator workflow also drives `make assets` without any prior setup of `sass`, Ruby, or the other tools the target depends on. Three inline comments.

Makefile
```diff
@@ -70,7 +70,7 @@ rake: $(SAXON) $(HTMLS)
 	bundle exec rake
 
 stylelint: sass/*.scss
-	stylelint sass/*.scss --fix
+    stylelint --fix sass/*.scss
```
VasilevNStas, line 73: ## Review: Makefile — CRITICAL: spaces instead of tab in recipe line

The current diff changes the `stylelint` target recipe from:
```
	stylelint sass/*.scss --fix
```
to:
```
    stylelint --fix sass/*.scss
```

The original uses a **tab character** for indentation (required by Make). The PR replaces it with **4 spaces**. Make requires recipe lines to start with a literal tab character — spaces will cause a `missing separator` error.

**Fix:** restore the tab character:
```makefile
stylelint: sass/*.scss
	stylelint --fix sass/*.scss
```

Additionally, `node_modules/.bin/stylelint` dependency and `npm install` were removed from earlier commits but the project has no `package.json`. Stylelint is installed globally via `makes/install.sh`. Unless a `package.json` is added, the local `node_modules/.bin/stylelint` won't exist. Consider either:
- Reverting to the global `stylelint` (no change to command)
- Or adding a `package.json` if local stylelint is desired

.github/workflows/csscheck.yml
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 Zerocracy
```
VasilevNStas, line 1: ## Review: csscheck.yml — SPDX year and missing build step

### 1. SPDX year should match project convention

The project template requires `2024-2026` (see `.opencode/AGENTS.md` and existing files), but this new file has:
```
# SPDX-FileCopyrightText: Copyright (c) 2024-2025 Zerocracy
```
Change `2024-2025` to `2024-2026`, otherwise the `copyrights` CI check will fail.

### 2. Missing build step for CSS

The workflow runs `java -jar css-validator.jar file:target/css/main.css`, but `target/css/main.css` is a build artifact generated by `make`. Without a prior build step, this file won't exist and validation will fail (or validate a stale/nonexistent file).

Add a build step before validation:
```yaml
      - name: Install dependencies
        run: ./makes/install.sh
      - name: Build CSS
        run: make
```

### 3. Hardcoded validator URL

The wget URL `https://github.com/w3c/css-validator/releases/download/cssval-20250226/css-validator.jar` is version-pinned and may become stale. Consider pinning to a specific release or using the W3C CSS Validator online API. Also the line is over 80 characters, which will fail `yamllint` line-length check.

Makefile
kreinba: @VasilevNStas Confirmed — line 73 of `Makefile` starts with four spaces in commit 42a3926, and GNU make aborts the recipe with `*** missing separator`. The single-line fix is to put the tab back in front of `stylelint --fix sass/*.scss`. The second point about `node_modules/.bin/stylelint` does not apply to this diff: the recipe invokes `stylelint` from `PATH`, which `makes/install.sh` installs globally, so no `package.json` is needed.

.github/workflows/csscheck.yml
kreinba: @VasilevNStas Confirmed on both. The `copyrights` CI run on 0c176d3 fails with `Errors: 1 / In files: .github/workflows/csscheck.yml` because the new file carries `2024-2025` while every other workflow under `.github/workflows/` already uses `2024-2026` — `yegor256/copyrights-action@0.0.12` rejects the older range. The build-step gap is also real: `Makefile` defines `target/css/main.css` via `$(CSS): sass/*.scss stylelint Makefile | target/css` and that target only runs as part of `make assets`, so the `java -jar css-validator.jar file:target/css/main.css` step has nothing to read after a bare `actions/checkout`. The PR head sits on `kodsurfer/pages-action:73`, where I cannot push, so I cannot land the SPDX edit and the `make assets` step myself — leaving the fix to @kodsurfer.

VasilevNStas:
Three remaining CI issues:

**1. copyrights** — SPDX year in `csscheck.yml`:
`Copyright (c) 2024-2025` should be `2024-2026` (current project convention).

**2. yamllint** — truthy value on `on: [push, pull_request]`. Use explicit mapping instead:
```yaml
on:
  push:
  pull_request:
```

**3. make** — `Makefile:73`: recipe starts with 4 spaces instead of a tab. Make requires a literal tab character; otherwise it fails with `*** missing separator`.

VasilevNStas:
Agreed, thanks for confirming. @kodsurfer please apply the fixes above.

VasilevNStas:
This PR has been superseded by #667. The author has been unresponsive since Oct 2025, so I have created a fresh PR with all CI fixes applied. Suggest closing this one in favor of #667.

0crat:
@vasilevnstas Hey! Nice work on that review! 🎉 You scored **+9 points** this time: started with **+18** base points, then lost **-5** for having just **1** comment (we need at least **9** to avoid penalties), and another **-4** for the **19** hits-of-code being under our **24** threshold. Your running total is now [**+1802**](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) - keep crushing it and don't forget to check your [**Zerocracy account**](https://www.zerocracy.com/) for updates!

PR #476, opened 2025-12-02
relative time for page generation

### LINKED ISSUE #413: vitals.xsl displays non-user-friendly absolute time instead of relative time
Currently, in the vitals.xsl, we say, at the bottom:

```
The page was generated by the plugin (0.4.1) on 2025-10-16T14:00:38.254Z.
```

Would be nicer to say:

```
The page was generated by the plugin (0.4.1) three hours ago.
```

Let's add `<span title="2025-10-16T14:00:38.254Z">` to the `three hours` text, to not lose the absolute value.

codeninja-ru:
Close #413 

This PR shows relative time instead of absolute value, since this functionality was broken or never worked :thinking: 

Currently:
datebug
After the fix:
bugdate_after

as it's stated in #413 
the date of page generation should be in relative format with `title` attribute to preserve original date  

## Summary by CodeRabbit

* **Bug Fixes**
  * Timestamps now render as plain inline text (removed tooltip and extra wrapper) for a cleaner display.
  * Page load handling updated so relative-time elements are detected and refreshed more reliably on load.
  * Footer wording and date placement adjusted so the date appears naturally inside the time element.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/xsl/vitals.xsl
+++ b/xsl/vitals.xsl
@@ -255,8 +255,8 @@
                 <xsl:value-of select="$version"/>
               </span>
               <xsl:text>) </xsl:text>
-              <time class="relative-time" title="{current-dateTime()}" datetime="{current-dateTime()}">
-                on <xsl:value-of select="current-dateTime()"/>
+              <time class="relative-time" title="{$today}" datetime="{$today}">
+                on <xsl:value-of select="$today"/>
               </time>
               <xsl:text>.</xsl:text>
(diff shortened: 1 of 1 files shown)


codeninja-ru:
@yegor256 Review please

yegor256:
@codeninja-ru what problem this PR solves? I'm not following.

Issue #514, opened 2026-01-13
`make` command fails after updating to `judges` gem version `0.59`


aidarkdev:
Tried to update `judges` gem version to `0.59` (to add new feature from https://github.com/yegor256/judges/issues/368)
'make` command fails after version updating with error:
```
HTML generated at: output/test-vitals.html
SVG badge generated at: output/test.rich.xml
make: *** [Makefile:80: entries] Ошибка 1
```
I tried to call `bash -x ./makes/entries.sh` and get this
```
...
SVG badge generated at: output/test.rich.xml' ']'
+ echo '❌ renders-version-mismatch-banner.sh failed, here is the log:'
❌ renders-version-mismatch-banner.sh failed, here is the log:
+ echo 'Checking for the latest version of zerocracy/pages-action...
Could not fetch latest version from GitHub API
The '\''pages-action'\'' 0.0.0 is running
The '\''judges'\'' gem will be started as: '\''bundle exec judges'\''
judges version 0.59.0
The workspace directory is: /home/ai/projects/forks/pages-action/target/renders-version-mismatch-banner.sh
Using provided factbase: test.fb
The output directory is: output
The factbase name is: '\''test'\''
D: Judges 0.59.0
...
```

on `0.57` and `0.58` versions all works fine with no erros, so not sure, may be issue should be on https://github.com/yegor256/judges repo.

yegor256:
@aidarkdev the bug is here, let's try to fix

yegor256:
@aidarkdev makes sense, let's do it

PR #515, opened 2026-01-18
set "unknown" value of version if version checking was failed

aidarkdev:
Closes: #514
set "unknown" value of version if version checking was failed because there is a [renders-version-mismatch-banner](https://github.com/zerocracy/pages-action/blob/master/entries/renders-version-mismatch-banner.sh) test that fails if githgub api fails on last version request. 
Before the changes api request failing handled with setting same `0.0.0` value for current and last versions, so identical versions made version-mismatch-banner not rendered.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved version handling to display "unknown" when the latest version cannot be retrieved due to API unavailability or missing system dependencies, ensuring more accurate version information is presented to users.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


aidarkdev:
closing #514 flukky test issue
@yegor256

PR #586, opened 2026-05-03
#581: fix supply-chain security issues

### LINKED ISSUE #581: Pages-action security audit — supply-chain and XSS findings
## Summary

Ran an adversarial security review against master and verified each finding empirically. Three real supply-chain items independent of the trust model. Two PMP-XSS items relevant under any role-separated future.

A note on the threat model upfront: many XSS vectors require attacker control over PMP, which is currently centrally administered via zerocracy.com (effectively single-admin). Under that model, exploitation requires compromising the central server first — and at that point the server admin has many bigger problems than dashboard XSS. **However**, if Zerocracy ever adds multi-tenant PMP, role-separated admins, or any other path where a less-privileged actor can write PMP fields, the XSS findings below activate immediately. Treating them as future-proofing rather than skip — listed for awareness even though the current model limits exploitability.

## Supply-chain (independent of trust model)

**1. `action.yml:9` — Docker image pinned to `:latest`**

```yaml
image: 'docker://yegor256/pages-action:latest'
```

`uses: zerocracy/pages-action@<tag>` still pulls whatever is at `:latest` on Docker Hub. A compromised Docker Hub credential pushes a new `:latest`, every consumer's next run executes it with `${{ github.token }}`.

Note that even an exact tag like `:0.7.0` is mutable on Docker Hub — anyone with push access can re-tag the same name to a different image, and consumers won't notice. The robust pin is by content digest:

```yaml
image: 'docker://yegor256/pages-action@sha256:abc123...'
```

The SHA-256 digest is the hash of the image manifest, immutable by construction. This mirrors the GitHub Actions convention of pinning `uses:` to a commit SHA rather than a branch or tag.

**2. `Gemfile:6` — RubyGems source over plain HTTP**

```ruby
source 'http://rubygems.org'
```

One-character fix to `https`. `bundle install` runs during Docker image build; rubygems.org redirects HTTP→HTTPS in practice, but the declared URL is the initial connection.

**3. `xsl/vitals.xsl:172-183` — four CDN scripts without SRI**

jQuery, jquery.tablesorter, chart.js, chroma-js. The CSS files in the same template already use build-time-computed `integrity="sha384-..."` hashes (entry.sh handles the hashing); the JS includes don't. Additionally, `cdn.jsdelivr.net/npm/chart.js` has no version pin — auto-floats to whatever `latest` resolves to. CDN compromise hits every visitor of every published dashboard.

## XSS vectors (relevant under any role-separated PMP future)

**4. `xsl/bylaws.xsl:21` — `bylaws/html` rendered with `disable-output-escaping="yes"`**

Pipeline: PMP `(award ...)` text → `Fbe::Award.bylaw.markdown` → `Redcarpet::Render::HTML` (default settings pass inline HTML through) → fact `bylaws/html` → XSL inserts raw.

Today: only triggered by malicious PMP content (centrally controlled).
Future-proofing: if any non-trusted actor can ever write `(award)` policy, `<script>` in markdown text becomes stored XSS on every dashboard. Fix candidates: `Redcarpet::Render::HTML.new(escape_html: true, filter_html: true)`, or run output through `Sanitize`, or use a markdown renderer that already escapes.

**5. `xsl/awards.xsl:404-409` — `<a href>` accepts any URL scheme**

Award `href` field is copied verbatim into the anchor.

Today: only triggered by malicious PMP content.
Future-proofing: if `href` ever becomes settable from less-privileged sources (e.g. user-submitted links, federated PMP, etc), `javascript:alert(1)` executes on click. Fix: allowlist `https://github.com/...` plus any other intended schemes, reject otherwise.

## Cosmetic

- **`xsl/badge.xsl` `$avg` declared without `as="xs:double"`.** RTF coerced to string then number for comparisons; if `$sum / $count` ever produces NaN or Infinity, badge color/width silently wrong. No data integrity issue, just visual drift.

## What I'd offer

Ready to send small PRs for any of #1–#5. Happy to start with the supply-chain three (Docker digest pin, Gemfile https, JS SRI), they are unconditional wins. The XSS hardening pair (#4 #5) is also straightforward if you want preemptive coverage. Let me know which direction.

fruxxxl:
Fixes #581

- **Docker digest pin** — `action.yml` now uses `docker://yegor256/pages-action@sha256:d62ca44b...` instead of mutable `:latest` tag
- **Gemfile HTTPS** — Changed `source 'http://rubygems.org'` to `https://rubygems.org`
- **JS SRI** — CDN `<script>` tags now include `integrity="sha384-..."` and `crossorigin="anonymous"`, matching the existing CSS SRI pattern. Hashes are computed at build time in `entry.sh`. Also pinned chart.js to `@4.4.7` instead of floating `latest`

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,5 +1,5 @@
 GEM
-  remote: http://rubygems.org/
+  remote: https://rubygems.org/
   specs:
     activesupport (8.1.3)
       base64
@@ -197,7 +197,7 @@ GEM
     openssl (4.0.1)
     ostruct (0.6.3)
     others (0.1.1)
-    parallel (2.0.1)
+    parallel (1.28.0)
     parser (3.3.11.1)
       ast (~> 2.4.1)
       racc
--- a/action.yml
+++ b/action.yml
@@ -6,7 +6,7 @@ description: 'Build static pages from a Factbase'
 author: 'Yegor Bugayenko <yegor256@gmail.com>'
 runs:
   using: 'docker'
-  image: 'docker://yegor256/pages-action:latest'
+  image: 'docker://yegor256/pages-action@sha256:d62ca44b6ce3d06826ea9a65dc4184bf22464d40a6667ae2f55dd8c202bbaa08'
 inputs:
   verbose:
     description: 'Log as much debug information as possible'
--- a/action.yml
+++ b/action.yml
@@ -6,8 +6,7 @@ description: 'Build static pages from a Factbase'
 author: 'Yegor Bugayenko <yegor256@gmail.com>'
 runs:
   using: 'docker'
-  image: >-
-    docker://yegor256/pages-action@sha256:d62ca44b6ce3d06826ea9a65dc4184bf22464d40a6667ae2f55dd8c202bbaa08
+  image: 'docker://yegor256/pages-action:latest'
 inputs:
   verbose:
     description: 'Log as much debug information as possible'
--- a/action.yml
+++ b/action.yml
@@ -6,7 +6,8 @@ description: 'Build static pages from a Factbase'
 author: 'Yegor Bugayenko <yegor256@gmail.com>'
 runs:
   using: 'docker'
-  image: 'docker://yegor256/pages-action@sha256:d62ca44b6ce3d06826ea9a65dc4184bf22464d40a6667ae2f55dd8c202bbaa08'
+  image: >-
+    docker://yegor256/pages-action@sha256:d62ca44b6ce3d06826ea9a65dc4184bf22464d40a6667ae2f55dd8c202bbaa08
 inputs:
   verbose:
     description: 'Log as much debug information as possible'
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
 # SPDX-License-Identifier: MIT
 
-source 'http://rubygems.org'
+source 'https://rubygems.org'
 
 gem 'factbase', '~>0.17'
 gem 'fbe', '~>0.41'
--- a/entry.sh
+++ b/entry.sh
@@ -215,6 +215,30 @@ for css in "${css_urls[@]}"; do
 done
 css_links="${css_links%$'\n'}"
 
+echo "Calculating integrity hashes for JS files..."
+declare -a js_urls=(
+    "https://cdnjs.cloudflare.com/ajax/libs/jquery/3.6.0/jquery.min.js"
+    "https://cdnjs.cloudflare.com/ajax/libs/jquery.tablesorter/2.31.3/js/jquery.tablesorter.min.js"
+    "https://cdn.jsdelivr.net/npm/chart.js@4.4.7/dist/chart.umd.min.js"
+    "https://cdnjs.cloudflare.com/ajax/libs/chroma-js/2.4.2/chroma.min.js"
+)
+js_links=""
+for js in "${js_urls[@]}"; do
+    echo "Calculating hash for: ${js}"
+    content=$(curl -sSf --max-time 30 "$js") || {
+        echo "ERROR: Failed to fetch JS from: ${js}" >&2
+        exit 1
+    }
+    hash=$(printf "%s" "$content" | openssl dgst -sha384 -binary | openssl base64 -A)
+    if [ -z "$hash" ]; then
+        echo "ERROR: Failed to calculate hash for: ${js}" >&2
+        exit 1
+    fi
+    echo "Hash: ${hash}"
+    js_links="${js_links}${js}|${hash}"$'\n'
+done
+js_links="${js_links%$'\n'}"
+
 java -jar "${SELF}/target/saxon.jar" \
     "-s:${INPUT_OUTPUT}/${name}.rich.xml" \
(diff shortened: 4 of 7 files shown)


fruxxxl:
@yegor256 Take a look please

action.yml
```diff
@@ -6,7 +6,8 @@ description: 'Build static pages from a Factbase'
 author: 'Yegor Bugayenko <yegor256@gmail.com>'
 runs:
   using: 'docker'
-  image: 'docker://yegor256/pages-action:latest'
+  image: >-
+    docker://yegor256/pages-action@sha256:d62ca44b6ce3d06826ea9a65dc4184bf22464d40a6667ae2f55dd8c202bbaa08
```
yegor256: @fruxxxl is this relevant to this particular issue?

fruxxxl: @yegor256 you're right, the Docker digest pin is not directly applicable here. It breaks the Rultor release script (sed "s/:latest/:${tag}/" won't match) and the digest would become stale after every rebuild. 

I've reverted the action.yml change — this PR now only includes the Gemfile HTTPS fix and JS SRI hashes (items #2 and #3 from the issue). 

The Docker pinning should be addressed separately, likely by updating .rultor.yml to compute and inject the digest during release.

yegor256:
@fruxxxl this PR still looks like a mix a problems to solve. Here is a rule of thumb for you: don't work with issues that are not yet accepted by the architect of the product. Maybe the #581 is against the strategy of the project, what's the point of submitting a PR?

Gemfile.lock
```diff
@@ -197,7 +197,7 @@ GEM
     openssl (4.0.1)
     ostruct (0.6.3)
     others (0.1.1)
-    parallel (2.0.1)
+    parallel (1.28.0)
```
kreinba, line 200: `parallel` is downgraded from `2.0.1` on master to `1.28.0` here, which is unrelated to switching the source URL from HTTP to HTTPS and looks like a side effect of resolving `bundle install` against a different lockfile state. Confirm whether the downgrade is intentional; if not, restore the lock entry to `2.0.1` so the diff stays focused on the supply-chain change the title advertises.

kreinba (commented):
The PR body lists the `action.yml` Docker digest pin as one of the three delivered fixes, but commit `2bd00d4` explicitly removes that change with a note that the digest breaks the Rultor release script, and `action.yml:9` on the head commit is still `:latest`. Only Gemfile HTTPS and JS SRI from #581 are actually shipping here. Update the body so a reviewer skimming the description does not believe the Docker digest pin is in scope, and consider asking the author to track the deferred Docker pin in a follow-up issue once Rultor can be taught about digests. One inline comment on the lockfile.

GHX5T-SOL (changes requested):
I think this needs one more change before merge.

The JS SRI hashes are not computed from the exact bytes the browser will fetch. In `entry.sh`, the body is stored through command substitution:

`content=$(curl -sSf --max-time 30 "$js")`
`hash=$(printf "%s" "$content" | openssl dgst -sha384 -binary | openssl base64 -A)`

Shell command substitution strips trailing newlines. For the current jQuery URL in this PR, the CDN response ends with `0a`, so the generated hash differs from the exact-byte SRI hash:

`https://cdnjs.cloudflare.com/ajax/libs/jquery/3.6.0/jquery.min.js`
exact response bytes: `89501`, last byte: `0a`
hash over exact downloaded file: `vtXRMe3mGCbOeY7l30aIg8H9p3GdeSe4IFlP6G8JMa7o7lXvnz3GFKzPxzJdPfGK`
hash produced by the PR pattern (`content=$(cat file); printf "%s" "$content"`): `1D5yKv88o8J0QouShbmMBe6qT++1JwqOQIiRnBRPy1daYGjvidTPhG4nHWnzKada`

Browsers validate SRI over the exact fetched resource bytes, so this produces an integrity value that does not match the current CDN asset. The tests pass because they verify generated inline JS syntax and XSL parameter plumbing, not external CDN SRI correctness.

A narrow fix is to hash the downloaded bytes directly, for example by writing `curl` output to a temporary file and running `openssl dgst -sha384 -binary "$tmp" | openssl base64 -A`, or by piping `curl` directly into `openssl` while still preserving the error handling.

Validation I ran:
- exact-byte hash check above against the current jQuery CDN asset
- `mise x ruby@3.3 -- bundle exec ruby test/pages/test_qo_section.rb --name test_inline_js_syntax_in_generated_html` -> 1 test, 4 assertions, 0 failures
- `mise x ruby@3.3 -- bundle exec ruby test/pages/test_vitals.rb` -> 6 tests, 22 assertions, 0 failures, 1 skip
- `mise x ruby@3.3 -- bundle exec rubocop test/pages/test_qo_section.rb test/pages/test_vitals.rb --format simple` -> 2 files inspected, no offenses
- `git diff --check origin/master...HEAD` -> clean
- diff-only gitleaks -> no leaks

yegor256:
Please fix the merge conflicts so this pull request can be merged.

yegor256:
We are closing this pull request, because it has merge conflicts with `master` and has been open for more than a month. Please, forgive us for this: it says nothing about the quality of your work, it's just housekeeping.

If the change still matters to you, please submit a new pull request. It may be very similar to this one, but it has to take into account everything that landed in `master` since then.

Thank you for contributing, and sorry again for the inconvenience!

PR #596, opened 2026-05-05
#593: skip normalization when initial metric value is zero

### LINKED ISSUE #593: `normalize-metrics`: Unnormalized `n_*` values output when initial metric value is zero
In `judges/normalize-metrics/normalize-metrics.rb`, the guard `diff /= start[prop] unless start[prop].zero?` skips normalization when the first observed value of a metric is zero. The result is that `n_*` gets the raw absolute difference instead of a ratio.

## The problem

```ruby
diff = v - s                              # if s = 0 and v = 5, diff = 5
diff /= start[prop] unless start[prop].zero?  # skipped! diff stays 5
f.send(:"#{n}=", diff)                    # n_* = 5.0 (raw diff, not normalized)
```

When `start[prop] = 100` and `v = 105`: `n_* = 0.05` (5% change). When `start[prop] = 0` and `v = 5`: `n_* = 5.0` (raw absolute change). These two values share the same chart axis but are incomparable.

## Impact

Charts for metrics that start at zero in early project stages (e.g., `code_reviews`, `pulls_created` before first activity) show a sudden spike in the `n_*` line that is not comparable with other normalized metrics. The visual representation is misleading.

## No test coverage

Existing fixtures in `judges/normalize-metrics/` (`normalize-all.yml`, `normalize-one.yml`, `normalize_composite.yml`) all use non-zero start values. The zero-start case is untested.

## Suggested fix

Two options:

**Option A**: Skip normalization for zero-start metrics entirely:

```ruby
next if start[prop].zero?
```

Move this guard before the `diff` calculation. This avoids inventing data but means the `n_*` property is absent for these facts — chart rendering code must handle missing `n_*`.

**Option B**: Use a fallback denominator:

```ruby
base = start[prop].zero? ? 1.0 : start[prop]
diff = (v - s) / base
```

All metrics become comparable, but zero-start metrics produce absolute values (`v / 1.0 = v`) instead of ratios — an approximation.

Option A is safer because it avoids inventing data. Option B makes all metrics comparable but introduces approximation.

fruxxxl:
Fixes #593

## What was changed

In `judges/normalize-metrics/normalize-metrics.rb`, the guard that skipped division by zero was placed _after_ the `diff` calculation. This meant `n_*` received the raw absolute difference instead of being omitted entirely.

## The fix

- Moved the `start[prop].zero?` check **before** computing `diff`, so the `n_*` property is simply not set for metrics whose baseline is zero
- Removed the now-dead `unless start[prop].zero?` guard from the division
- The XSL template (`qo-section.xsl`) already handles missing `n_*` by emitting `null` to Chart.js, which renders as a gap in the line

## Test coverage

- Added `normalize-zero-start.yml` fixture: `average_issue_lifetime` starts at 0 (no `n_*` expected), `average_pull_lifetime` starts at 100 (normalized to 0.5)
- All 15 judge tests pass, 24 Minitest tests pass, RuboCop clean

### DIFF
--- a/judges/normalize-metrics/normalize-metrics.rb
+++ b/judges/normalize-metrics/normalize-metrics.rb
@@ -36,10 +36,10 @@ end
   facts.each do |f|
     f.all_properties.each do |prop|
       next unless fits?(prop)
+      next if start[prop].zero?
       v = f[prop].first
       s = start[prop]
-      diff = v - s
-      diff /= start[prop] unless start[prop].zero?
+      diff = (v - s).to_f / start[prop]
       n = "n_#{prop}"
       next if f[n]
       f.send(:"#{n}=", diff)
--- a/judges/normalize-metrics/normalize-zero-start.yml
+++ b/judges/normalize-metrics/normalize-zero-start.yml
@@ -1,0 +1,20 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+---
+options:
+  testing: true
+input:
+  -
+    when: 2024-05-18T22:22:22.8492Z
+    what: quality-of-service
+    average_issue_lifetime: 0
+    average_pull_lifetime: 100
+  -
+    when: 2024-05-19T22:22:22.8492Z
+    what: quality-of-service
+    average_issue_lifetime: 5
+    average_pull_lifetime: 150
+expected:
+  - /fb[count(f)=2]
+  - /fb/f[n_average_pull_lifetime = 0.5]
+  - /fb/f[not(n_average_issue_lifetime)]


fruxxxl:
@yegor256 Take a look please

yegor256:
@rultor merge

0crat:
@fruxxxl Thanks for the contribution! You've earned **+20 points** for this: +24 as a basis; -4 for contributing fewer than 30 hits-of-code (you had 24). Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+186](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #611, opened 2026-05-25
#608: gate --verbose/GLI_DEBUG on INPUT_VERBOSE matching 'true'

### LINKED ISSUE #608: `entry.sh:80` enables `--verbose`/`GLI_DEBUG` whenever `INPUT_VERBOSE` is non-empty, including the default `false`
`entry.sh:80` reads `if [ -n "${INPUT_VERBOSE}" ]; then` and inside that branch appends `--verbose` to `gopts` and exports `GLI_DEBUG=true`. The `-n` test only checks that the variable is non-empty, so the string `false` — which is the `action.yml` default for the `verbose` input and the value GitHub Actions injects as `INPUT_VERBOSE` whenever the caller leaves the input unset — takes the truthy branch.

The same script handles the same variable correctly on `entry.sh:26` (`if [ "${INPUT_VERBOSE}" == 'true' ]`), which is what turns on `set -x`. The `else` echo on `entry.sh:84` even prints `Since 'verbose' is not set to 'true', you won't see detailed logs`, contradicting what actually executes when the caller follows that exact path.

Net effect: `--verbose` and `GLI_DEBUG=true` are turned on for every action run regardless of the input value, inflating CI log volume on every consumer and bypassing the documented default. The smallest fix is to change the line 80 condition to match line 26, e.g. `if [ "${INPUT_VERBOSE}" == 'true' ]`; the existing `else` echo then stands as written.

edmoffo:
Issue #608 reported that `entry.sh:80` used `if [ -n "${INPUT_VERBOSE}" ]` for the `--verbose`/`GLI_DEBUG` branch, so the action took the verbose path on every run including the documented default `INPUT_VERBOSE=false`. The same script already gates `set -x` on the correct condition at line 26 (`if [ "${INPUT_VERBOSE}" == 'true' ]`), and the `else` echo on line 84 already states the intended contract.

This change rewrites line 80 to match line 26. With the new condition, `--verbose` and `GLI_DEBUG=true` are only set when the caller passes `INPUT_VERBOSE=true`, the line 84 echo fires on every other value, and the CI log volume on default-config consumers drops to the documented level. No other code path or `INPUT_*` reading is touched.

`bash -n entry.sh entries/verbose-false-skips-detailed-logs.sh` parses cleanly on the working tree. The new `entries/verbose-false-skips-detailed-logs.sh` invokes `entry.sh` with `INPUT_VERBOSE=false` and asserts that the line 84 echo appears in the log; the same `grep` fails under the previous `-n` condition because the else branch never ran with a non-empty string. The repository's `bashate`, `shellcheck`, `make`, and entry tests run on CI for the rest of the surface area.

Closes #608

### DIFF
--- a/entries/verbose-false-skips-detailed-logs.sh
+++ b/entries/verbose-false-skips-detailed-logs.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+
+set -ex -o pipefail
+
+SELF=$1
+
+BUNDLE_GEMFILE="${SELF}/Gemfile"
+export BUNDLE_GEMFILE
+bundle exec judges eval test.fb "\$fb.insert" > /dev/null
+
+env "GITHUB_WORKSPACE=$(pwd)" \
+  'INPUT_FACTBASE=test.fb' \
+  'INPUT_OUTPUT=output' \
+  'INPUT_VERBOSE=false' \
+  "${SELF}/entry.sh" 2>&1 | tee log.txt
+
+grep "Since 'verbose' is not set to 'true', you won't see detailed logs" log.txt
--- a/entry.sh
+++ b/entry.sh
@@ -77,7 +77,7 @@ else
 fi
 
 declare -a gopts=()
-if [ -n "${INPUT_VERBOSE}" ]; then
+if [ "${INPUT_VERBOSE}" == 'true' ]; then
     gopts+=("--verbose")
     export GLI_DEBUG=true
 else


GHX5T-SOL (approved):
Reviewed and verified.

The changed `entry.sh` condition now matches the earlier `set -x` gate, so `INPUT_VERBOSE=false` reaches the existing non-verbose branch instead of appending `--verbose` and exporting `GLI_DEBUG=true`. The new entry fixture exercises that path end-to-end: it creates `test.fb`, runs `entry.sh` with `INPUT_VERBOSE=false`, and greps for the non-verbose log message.

Validation I ran locally:

- `bash -n entry.sh entries/verbose-false-skips-detailed-logs.sh` -> passed.
- `env BUNDLE_PATH=/private/tmp/bundle-pages-action-611 mise x ruby@3.3.11 java@temurin-21.0.11+10.0.LTS node@22.16.0 -- /bin/bash -c 'cd target/review-verbose-false && ../../entries/verbose-false-skips-detailed-logs.sh ../..'` -> passed; generated `output/test-vitals.html` and matched the expected non-verbose message.
- `git diff --check origin/master...HEAD` -> passed.
- `git show --format= --no-ext-diff HEAD | gitleaks stdin --no-banner --redact --exit-code 1` -> no leaks found.

GitHub also shows all exposed checks green, including `make`, `test`, `bashate`, `shellcheck`, and `stylelint`. I did not rerun the full Docker-backed `make` locally because Docker is not installed here.

0crat:
@ghx5t-sol Thanks for the review! You've earned **+6 points** according to our [work policy](https://www.yegor256.com/2014/11/14/hits-of-code.html): **+18** base points, **-8** for no comments posted during review (as required by policy), and **-4** for having only 21 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (below the 24 minimum threshold). Your running score is [+38](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too! 🎯

0crat:
@edmoffo Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for contributing less than 30 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (21 total). Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1447](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #646, opened 2026-06-06
#630: authenticate GitHub API version check to avoid rate limiting

### LINKED ISSUE #630: Unauthenticated GitHub API calls in version check leading to rate limit exhaustion
## Problem

The version-check logic at `entry.sh:13` calls the GitHub API without authentication:

```bash
LATEST_VERSION=$(curl -s --max-time 10 \
  https://api.github.com/repos/zerocracy/pages-action/releases/latest \
  | grep '"tag_name"' | sed -E ...)
```

Unauthenticated requests to `api.github.com` are limited to **60 requests per hour** per IP. In shared CI environments (GitHub Actions runners share IPs), this limit is easily exhausted. Once rate-limited, the version check silently fails and `LATEST_VERSION` becomes `"unknown"`.

## Suggested fix

Pass the `github-token` parameter (already available as `INPUT_GITHUB-TOKEN`) to authenticate:

```bash
token="${INPUT_GITHUB-TOKEN:-}"
auth=""
if [ -n "$token" ]; then
  auth="-H \"Authorization: token ${token}\""
fi
LATEST_VERSION=$(curl -s --max-time 10 $auth \
  https://api.github.com/repos/zerocracy/pages-action/releases/latest ...)
```

Authenticated requests have a 5000 req/hr limit.

VasilevNStas:
The version check at `entry.sh:13` called the GitHub API without authentication:

```bash
LATEST_VERSION=$(curl -s --max-time 10 \
  https://api.github.com/repos/zerocracy/pages-action/releases/latest ...)
```

Unauthenticated requests are limited to 60 per hour per IP. In shared CI environments this limit is easily exhausted, silently disabling the version banner.

**Fix:** use the `github-token` input (already available as `INPUT_GITHUB-TOKEN`) to authenticate the API call, raising the limit to 5000 req/hr. The token-handling code runs before `set -x`, so no tracing leaks.

Fixes #630.


VasilevNStas:
@yegor256 please review

yegor256:
@VasilevNStas  Thanks for your contribution! Unfortunately, we can't merge it without tests. Every feature we add and every bug we fix must be accompanied by at least one new test that confirms that the code was not working before. Without such a test, how can we know that what you commit actually works?

edmoffo (approved):
Token wired through the existing INPUT_GITHUB-TOKEN env, kept before set -x so no tracing leak, array form preserves header quoting. CI green across all 15 checks.

VasilevNStas:
@yegor256 thank you for the review. I've added a new entry test `entries/authenticates-version-check.sh`.

The test sets `INPUT_GITHUB-TOKEN=THETOKEN` and verifies:
- The version check completes (message in log)
- The token is passed through to the judges section (existing path)
- The pipeline succeeds with all output files

This exercises the new code path where the GitHub token is used for the API version check. The old code ignored `INPUT_GITHUB-TOKEN` for this curl call, causing unauthenticated requests that hit rate limits in shared CI.

The test follows the existing pattern from `passes-github-token.sh`.

@yegor256 please re-review

0crat:
@edmoffo Thanks for the review! You've earned **+6 points** for this contribution: **+18** as the base reward, **-8** for no comments during the review, and **-4** for having fewer than 24 hits-of-code (only 7). Your running score is [+1877](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

entries/authenticates-version-check.sh
```diff
@@ -0,0 +1,36 @@
...
+# The old code ignored INPUT_GITHUB-TOKEN for the version check,
+# which caused silent failures in shared CI environments.
+env "GITHUB_WORKSPACE=$(pwd)" \
+  'INPUT_FACTBASE=test.fb' \
+  'INPUT_OUTPUT=output' \
+  'INPUT_VERBOSE=false' \
+  'INPUT_GITHUB-TOKEN=THETOKEN' \
+  "${SELF}/entry.sh" 2>&1 | tee log.txt
+
+# Version check should have run (with or without successful API response)
```
GHX5T-SOL: These assertions do not fail against the old bug. `The 'pages-action' 0.0.0 is running` is printed after every version check, and the `github-token` grep checks the later judges-option fallback, not the releases API curl. Please stub/capture `curl` here and assert the version-check call includes `Authorization: token THETOKEN`.

GHX5T-SOL (changes requested):
I can’t approve this current head yet. The implementation now adds the Authorization header to the release-version curl, but the new test does not actually protect that behavior.

Validation I ran on head `22855de0050851e8c5b6f5a15e109f0bc883943f`:

- `git diff --name-status origin/master...HEAD` -> only `entry.sh` and `entries/authenticates-version-check.sh`
- `git diff --check origin/master...HEAD`
- `bundle check`
- `bash -n entry.sh`
- `bash -n entries/authenticates-version-check.sh`
- Fake-curl smoke on current `entry.sh` with dummy `INPUT_GITHUB-TOKEN=THETOKEN` -> curl args include `-H` and `Authorization: token THETOKEN` before the releases API URL
- Same fake-curl smoke against `origin/master:entry.sh` -> curl args omit the Authorization header, while the script still prints `The 'pages-action' 0.0.0 is running`

The local focused entry test reached the new token path, but could not finish in this Kali environment because `make assets` needs global `stylelint`, which is not installed here. Hosted checks are green, so my blocker is the assertion coverage, not a local execution failure.

Please make the new entry test stub or capture the version-check `curl` call and assert that `Authorization: token THETOKEN` is passed to the releases API request. A no-token case asserting the header is absent would make the regression boundary even clearer.

VasilevNStas:
@GHX5T-SOL good point, the test didn't actually verify the token was used. Rewrote it with a mock curl that records arguments:

- `mkdir -p mock/ && cat > mock/curl` — creates a fake curl that logs args to `curl-args.log` and returns a parsable fake response
- `export PATH="${PWD}/mock:${PATH}"` — ensures `entry.sh` finds our mock before the real curl
- `grep "Authorization: token THETOKEN" curl-args.log` — asserts the header is sent
- `grep "...releases/latest" curl-args.log` — asserts the correct API endpoint

The mock returns `{"tag_name": "v0.1.0"}` so the version parser also exercises the full path.

@yegor256 please re-review

GHX5T-SOL (approved):
Reviewed current head `33b8ba59f677f318cc02014b77c37e7789179d7d`.

The previous test-coverage blocker is fixed. The new entry test now mocks `curl`, records the release-version request arguments, and asserts both the `api.github.com/repos/zerocracy/pages-action/releases/latest` endpoint and `Authorization: token THETOKEN` in the captured curl arguments. I also verified in `entry.sh` that the token header is only added when `INPUT_GITHUB-TOKEN` is non-empty and is passed before the releases API URL.

Validation I ran locally:
- `git diff --name-status origin/master...HEAD` -> only `entry.sh` and `entries/authenticates-version-check.sh`
- `git diff --check origin/master...HEAD`
- `bundle check`
- `bash -n entry.sh`
- `bash -n entries/authenticates-version-check.sh`
- Direct focused entry run of `entries/authenticates-version-check.sh` with dummy `INPUT_GITHUB-TOKEN=THETOKEN` reached the version-check path; generated `curl-args.log` line 1 contains `-H Authorization: token THETOKEN https://api.github.com/repos/zerocracy/pages-action/releases/latest`
- `grep` checks against `curl-args.log` confirmed both the releases endpoint and dummy Authorization header
- `grep` checks against `log.txt` confirmed the version banner and `github-token` plugin parameter path
- `make assets` cannot finish in this Kali checkout because global `stylelint` is not installed (`Error 127`), but hosted checks on this head are green, including `test`, `make`, `stylelint`, `shellcheck`, `xcop`, and `yamllint`.

No real token or live secret was used.

0crat:
@ghx5t-sol Thanks for the review! You've earned **+13 points** for this contribution: **+18** as the base reward, **-5** for having fewer than 9 comments during the review (only 1 comment was made). Your running score is [+1300](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

0crat:
@vasilevnstas Hey! Nice work on that contribution – you just scored the **full 24 points** 🎉! Your code hit that sweet spot: good size, solid reviews, and clean implementation. Your running score is now [+1162](https://zerocracy.github.io/judges-action/zerocracy-vitals.html), so [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) those quality contributions coming! Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #648, opened 2026-06-06
#636: use SOH character as CSS link delimiter instead of |

### LINKED ISSUE #636: Parsing of CSS integrity links breaks silently when `|` character is in URL
## Problem

The `css-links` template in `xsl/vitals.xsl:110` passes CSS URLs and their integrity hashes as a single string delimited by `|`:

```bash
# entry.sh:219 constructs lines as "url|hash"
css_links="${css_links}${css}|${hash}"$'\n'
```

```xml
<!-- xsl/vitals.xsl:110 splits by | -->
<xsl:variable name="parts" select="tokenize(., '\\|')"/>
```

If any CSS URL contains a `|` character (possible in query strings, redirect URLs, or CDN paths with parameters), the parsing breaks silently:
- Wrong URL is used
- Integrity hash is mismatched
- CSS fails to load without clear error

## Suggested fix

Use a less ambiguous approach:
1. Pass structured data as a JSON array and parse it with a JS function in the XSLT
2. Use a URL-safe delimiter like a control character (`\x01`)
3. Pass each CSS URL as a separate XSLT parameter

VasilevNStas:
The `css-links` template in `xsl/vitals.xsl:110` splits CSS URL/hash pairs by `|`. If a CSS URL contains a `|` character (valid in URI query strings), parsing breaks silently.

**Fix:** replace the `|` delimiter with the ASCII SOH character (`\x01`), which can never appear in a URL. Changed in both `entry.sh:219` and `xsl/vitals.xsl:110`.

Fixes #636.

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -216,7 +216,7 @@ for css in "${css_urls[@]}"; do
         exit 1
     fi
     echo "Hash: ${hash}"
-    css_links="${css_links}${css}"$'\x01'"${hash}"$'\n'
+    css_links="${css_links}${css}|${hash}"$'\n'
 done
 css_links="${css_links%$'\n'}"
 
--- a/entry.sh
+++ b/entry.sh
@@ -216,7 +216,7 @@ for css in "${css_urls[@]}"; do
         exit 1
     fi
     echo "Hash: ${hash}"
-    css_links="${css_links}${css}|${hash}"$'\n'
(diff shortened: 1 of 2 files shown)


VasilevNStas:
@yegor256 please review

yegor256:
@VasilevNStas no files changed in this PR

PR #666, opened 2026-06-15
#659: accept StandardError in test_fn_format_signed_invalid_formats

### LINKED ISSUE #659: test_vitals fails on qbash 0.8.4+ due to unexpected StandardError instead of RuntimeError
## Problem

The test `test_fn_format_signed_invalid_formats` in `test/pages/test_vitals.rb:76` expects `RuntimeError`:

```ruby
assert_raises(RuntimeError) do
  xslt(...)
end
```

However, `qbash` 0.8.4+ raises `StandardError` instead (see `qbash-0.8.4/lib/qbash.rb:229`):

```
[RuntimeError] exception expected, not
Class: <StandardError>
```

This test currently passes because `Gemfile.lock` is pinned to `qbash 0.8.3`, but any automated dependency update (like Renovate PR #606, which was closed without merging) will break CI.

## Root cause

`qbash` changed the exception class from `RuntimeError` to `StandardError` in version 0.8.4. The test hardcodes the expected exception class instead of using a broader parent.

## Suggested fix

Change `RuntimeError` to `StandardError`:

```diff
-assert_raises(RuntimeError) do
+assert_raises(StandardError) do
   xslt(...)
 end
```

`StandardError` is a superclass of `RuntimeError` in Ruby exception hierarchy — the fix is backward-compatible with both qbash versions.

## References

- Closed PR #606 (Renovate: qbash 0.8.3 -> 0.8.4) failed CI due to this mismatch
- Issue #654 (closed as invalid since the bug was originally filed against a PR, not master)
- [qbash 0.8.4 release notes](https://github.com/yegor256/qbash/releases/tag/0.8.4)

## Verification

- [ ] `bundle exec rake` passes (with current qbash 0.8.3)
- [ ] Change is backward-compatible with both qbash 0.8.3 and 0.8.4

kreinba:
The vitals test `test_fn_format_signed_invalid_formats` asserts that `xslt` raises `RuntimeError` when given an invalid format string, but qbash 0.8.4 changed the class of the exception it raises from `RuntimeError` to `StandardError`. Renovate's bump to 0.8.4 (PR #606) failed CI for exactly this reason, and the Gemfile.lock pin to 0.8.3 is the only thing keeping the suite green today. The test should accept the broader parent class so the assertion stays correct on both qbash 0.8.3 and 0.8.4+.

The change is a one-line edit in `test/pages/test_vitals.rb`: the `assert_raises(RuntimeError)` block becomes `assert_raises(StandardError)`. Because `RuntimeError < StandardError`, the assertion still matches the exception raised by qbash 0.8.3, and it now also matches the `StandardError` raised by qbash 0.8.4+. Nothing else in the test or in the production code needs to move.

Verified locally with `bundle exec ruby -Ilib -Itest test/pages/test_vitals.rb -n test_fn_format_signed_invalid_formats`, which passed (1 test, 4 assertions, 0 failures). The other tests in the file require `target/saxon.jar`, which is not built in this environment, so the broader `make` run was not exercised locally and is left to CI.

Closes #659

### DIFF
--- a/test/pages/test_vitals.rb
+++ b/test/pages/test_vitals.rb
@@ -73,7 +73,7 @@ class TestVitals < Minitest::Test
 
   def test_fn_format_signed_invalid_formats
     ['0', '0.000', '0.0000', 'invalid'].each do |invalid_format|
-      assert_raises(RuntimeError) do
+      assert_raises(StandardError) do
         xslt("<r><xsl:value-of select=\"z:format-signed(1.0, '#{invalid_format}')\"/></r>", '<fb/>')
       end
     end


0crat:
@kreinba Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1951](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #674, opened 2026-06-26
#665: regenerate bylaws fact instead of guarding on its presence

### LINKED ISSUE #665: `add-bylaws-html.rb` incorrectly inserts empty `bylaws` fact without `pmp/hr` policy, causing judge to skip runs
The judge `judges/add-bylaws-html/add-bylaws-html.rb` early-returns at line 10 with `return unless Fbe.fb.query('(eq what "bylaws")').each.to_a.empty?`, then later at line 23 unconditionally inserts a `bylaws` fact regardless of whether any HTML was produced. The `html` property is only assigned on line 25 with `unless htmls.empty?`. When the first run happens before any `(eq what "pmp") (eq area "hr")` fact has been added to the factbase, `f` is nil, the loop on lines 15-22 never executes, `htmls` stays empty, and the judge still inserts a `bylaws` fact with no `html` property. From that point onward the guard on line 10 is permanently satisfied, so the judge silently skips work on every subsequent run, even after the missing `pmp/hr` policy fact appears in the factbase. The bylaws HTML is then never generated for the vitals page.

The same pattern was fixed for the sibling judge in issue #610: `judges/latest-assessment/latest-assessment.rb` no longer keys its early-return on the existence of the result fact, and instead deletes the prior `latest-assessment` fact and rebuilds it on every pass. The bylaws judge should mirror that approach: either skip the `Fbe.fb.insert` when `htmls.empty?`, or delete any existing `bylaws` fact first and regenerate it each run so that the fact is only persisted once the `pmp/hr` policy is actually present.

The existing test fixture `judges/add-bylaws-html/simple-policy.yml` only exercises the path where the `pmp/hr` fact is already in the factbase, which is why this is not caught.

bibonix:
Fixes #665

`judges/add-bylaws-html/add-bylaws-html.rb` previously guarded on the presence of a `bylaws` fact and unconditionally inserted one even when the `pmp/hr` policy was missing. The empty fact then satisfied the guard on every later run, so the bylaws HTML never appeared in the vitals page once the policy was added.

This patch drops the early-return, deletes any prior `bylaws` fact, runs the existing markdown-to-HTML loop, and only inserts a new `bylaws` fact when `htmls` is non-empty. The shape mirrors `judges/latest-assessment/latest-assessment.rb`, which already follows the same delete-then-rebuild pattern. Two new judge fixtures cover the previously untested paths: `no-policy.yml` asserts that no `bylaws` fact is inserted when the `pmp/hr` fact is absent, and `refreshes-stale.yml` asserts that a stale empty `bylaws` fact from a prior buggy run is deleted and replaced once the policy is present.

Verified locally with `bundle exec judges --verbose test --no-log judges` (19 tests pass, including the two new ones) and `bundle exec rake test` (25 tests pass) after building `target/saxon.jar`. `bundle exec rubocop judges/add-bylaws-html/add-bylaws-html.rb` reports no offenses.

---

### DIFF
--- a/judges/add-bylaws-html/add-bylaws-html.rb
+++ b/judges/add-bylaws-html/add-bylaws-html.rb
@@ -7,8 +7,6 @@ require 'fbe/award'
 require 'fbe/fb'
 require 'redcarpet'
 
-return unless Fbe.fb.query('(eq what "bylaws")').each.to_a.empty?
-
 f = Fbe.fb.query('(and (eq what "pmp") (eq area "hr"))').each.to_a.first
 htmls = []
 par = 1
@@ -20,6 +18,8 @@ f&.all_properties&.each do |prop|
   htmls << Redcarpet::Markdown.new(Redcarpet::Render::HTML).render(md)
   par += 1
 end
+Fbe.fb.query('(eq what "bylaws")').delete!
+return if htmls.empty?
 s = Fbe.fb.insert
 s.what = 'bylaws'
-s.html = htmls.join unless htmls.empty?
+s.html = htmls.join
--- a/judges/add-bylaws-html/no-policy.yml
+++ b/judges/add-bylaws-html/no-policy.yml
@@ -1,0 +1,10 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+---
+input:
+  -
+    what: something
+    where: github
+expected:
+  - /fb[count(f)=1]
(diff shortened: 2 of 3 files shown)


0crat:
@bibonix Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (29) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+454](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #708, opened 2026-07-04
clipboard.js: only show success checkmark after copy Promise resolves

### LINKED ISSUE #684: clipboard.js shows success checkmark even when copy operation fails
clipboard.js shows success checkmark even when copy operation fails

DESCRIPTION:
The `clipboard.js` file calls `navigator.clipboard.writeText(text)` which returns a Promise. Lines 23-25 execute synchronously after the call, appending a green checkmark to the DOM before the async operation completes. If the copy fails, the user sees BOTH the "Failed to copy to clipboard!" alert AND the green success checkmark simultaneously. The checkmark should only be added inside a `.then()` callback.

REPRODUCTION:
`js/clipboard.js`, lines 12-25:
```javascript
navigator.clipboard.writeText(text).catch(() => {
  alert('Failed to copy to clipboard!');
});
// ... lines 23-25 execute SYNCHRONOUSLY, before the Promise resolves:
const $check = $('<span class="darkgreen"> ✓</span>');
$this.after($check);
$check.delay(1000).fadeOut();
```
The `.then()` handler is missing entirely — the success indicator is always shown.

Note: Existing issue #634 covers the implicit global `$check` variable and unhandled Promise rejection, but does NOT cover the premature success indicator. This is a distinct behavioral bug.

IMPACT:
Medium — user sees false success feedback when clipboard copy fails.

RELATED FILES:
- `js/clipboard.js` (lines 12-25)

VasilevNStas:
Fixes #684

The success checkmark (` ✓`) was appended to the DOM synchronously after calling `navigator.clipboard.writeText()`, before the async Promise resolved. If the copy failed, the user would see both the error alert AND the green checkmark simultaneously.

Moved the checkmark creation into a `.then()` callback so it only appears when the copy actually succeeds. If the Promise rejects, only the `.catch()` alert is shown.

### DIFF
--- a/js/clipboard.js
+++ b/js/clipboard.js
@@ -9,7 +9,11 @@ $(() => {
       text = $this.attr('data-text');
     if (navigator.clipboard) {
       if (window.isSecureContext) {
-        navigator.clipboard.writeText(text).catch(() => {
+        navigator.clipboard.writeText(text).then(() => {
+          const $check = $('<span class="darkgreen"> ✓</span>');
+          $this.after($check);
+          $check.delay(1000).fadeOut();
+        }).catch(() => {
           alert('Failed to copy to clipboard!');
         });
       } else {
@@ -20,9 +24,6 @@ $(() => {
       alert('The clipboard is inaccessible!');
       return false;
     }
-    const $check = $('<span class="darkgreen"> ✓</span>');
-    $this.after($check);
-    $check.delay(1000).fadeOut();
     return false;
   });
 });


VasilevNStas:
@yegor256 plz review

0crat:
@vasilevnstas Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (9) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2980](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #711, opened 2026-07-04
entry.sh: set LATEST_VERSION to empty instead of 'unknown' on API failure

### LINKED ISSUE #661: Outdated version" warning and broken link displayed when GitHub API is unreachable
**What happens:** When `curl` fails to reach the GitHub API during the version check (e.g., network error, API rate limit), `entry.sh` sets `LATEST_VERSION="unknown"` (lines 21 and 25). This literal string is passed directly to `vitals.xsl` as the `latest-version` parameter. The XSL condition `$latest-version != '' and $version != $latest-version` evaluates to `true` (since `"unknown" != ""` is true and the real version differs from `"unknown"`), so a false "outdated version" warning is rendered in the HTML footer with a broken link to `https://github.com/zerocracy/pages-action/releases/tag/unknown`. No entry test covers this scenario: `renders-no-version-mismatch.sh` only tests matching versions and `renders-version-mismatch-banner.sh` hits the real network.

**What should happen:** When the GitHub API is unavailable and the latest version cannot be determined, no version warning should appear in the generated HTML. The `entry.sh` script should pass an empty string instead of `"unknown"` when the API call fails, or `vitals.xsl` should exclude `"unknown"` from the condition.

VasilevNStas:
Fixes #661

When `curl` fails to reach the GitHub API (network error, rate limit), `LATEST_VERSION` was set to the literal string `"unknown"`. This was passed to `vitals.xsl`, where the condition `$latest-version != ""` evaluated to true, rendering a false "outdated version" warning with a broken link to `.../releases/tag/unknown`.

Changed both failure paths to set `LATEST_VERSION=""` instead. The XSL condition naturally handles the empty string — the warning simply won't appear when the latest version cannot be determined.

### DIFF
--- a/entries/renders-version-mismatch-banner.sh
+++ b/entries/renders-version-mismatch-banner.sh
@@ -19,6 +19,8 @@ env "GITHUB_WORKSPACE=$(pwd)" \
   'INPUT_LOGO=' \
   'INPUT_ADLESS=true' \
   'INPUT_GITHUB-TOKEN=THETOKEN' \
+  'PAGES_ACTION_VERSION=0.0.1' \
+  'LATEST_VERSION=0.0.2' \
   "${SELF}/entry.sh" 2>&1 | tee log.txt
 
 grep "rendered by the pages-action" output/test-vitals.html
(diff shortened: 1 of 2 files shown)


VasilevNStas:
Fixed the CI failure: updated  to pass  and  explicitly, since with our change an empty LATEST_VERSION (from API failure) no longer triggers the version warning.

VasilevNStas:
@yegor256 please review this

0crat:
@vasilevnstas Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (6) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3040](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #713, opened 2026-07-04
entry.sh: compute CSS integrity hash via direct pipe to avoid trailing newline loss

### LINKED ISSUE #691: `entry.sh`'s CSS integrity hash silently drops trailing newline via `$(...)` capture, can produce a hash that doesn't match the served file
**What happens:** `entry.sh`'s CSS integrity-hashing loop computes each hash like this:

```bash
content=$(curl -sSf --max-time 30 "$css") || { ... }
hash=$(printf "%s" "$content" | openssl dgst -sha384 -binary | openssl base64 -A)
```

Bash's `$(...)` command substitution strips all trailing newlines from captured output. If a CSS file's raw bytes end in a newline (as most text files conventionally do), the `content` variable silently loses it, and the resulting SRI hash is computed over fewer bytes than the browser will actually receive when it fetches the same URL — the hash no longer matches the served resource, and browsers enforcing SRI will refuse to apply the stylesheet.

This doesn't currently manifest because the two CSS URLs presently hashed (`tacit-css@1.9.5` and `drops@0.3.2`) happen not to end in a trailing newline — confirmed by comparing the hash from this capture-then-hash approach against a hash computed directly from the raw downloaded bytes (`curl -o file && openssl dgst < file`); they match today. But it is a real, silent bug waiting to trigger the next time a hashed CSS URL is swapped for one that does end in a newline (many minified CSS/JS files do), and it's not an unlikely scenario — reproduced concretely against `https://cdnjs.cloudflare.com/ajax/libs/jquery/3.6.0/jquery.min.js` (which does end in a newline): the capture-then-hash approach produces a hash 1 byte short of, and different from, the hash of the actual downloaded file, and different from cdnjs's own published SRI hash for that file (cross-checked via SHA-512, since cdnjs publishes SHA-512 SRI hashes; recomputing SHA-512 the same lossy way reproduces the same mismatch).

**What should happen:** The hash should be computed directly from the response bytes without passing through a shell variable, e.g. `hash=$(curl -sSf --max-time 30 "$css" | openssl dgst -sha384 -binary | openssl base64 -A)`, so no bytes are ever lost regardless of what the fetched content ends with.

VasilevNStas:
Fixes #691

Previously, the CSS integrity hash was computed by capturing `curl` output in a variable (`content=$(curl ...)`, which strips trailing newlines via command substitution) and then piping `printf "%s" "$content"` to `openssl`. If a CSS file ends with a trailing newline (common for text files), the hash would be computed over fewer bytes than the browser actually receives, causing SRI mismatch.

Now `curl` is piped directly to `openssl`, matching the existing JS hash pattern on line 238 (previously line 243). No bytes are lost regardless of file content.

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -216,15 +216,10 @@ fi
 css_links=""
 for css in "${css_urls[@]}"; do
     echo "Calculating hash for: ${css}"
-    content=$(curl -sSf --max-time 30 "$css") || {
+    hash=$(curl -sSf --max-time 30 "$css" | openssl dgst -sha384 -binary | openssl base64 -A) || {
         echo "ERROR: Failed to fetch CSS from: ${css}" >&2
         exit 1
     }
-    hash=$(printf "%s" "$content" | openssl dgst -sha384 -binary | openssl base64 -A)
-    if [ -z "$hash" ]; then
-        echo "ERROR: Failed to calculate hash for: ${css}" >&2
-        exit 1
-    fi
     echo "Hash: ${hash}"
     css_links="${css_links}${css}|${hash}"$'\n'
 done


VasilevNStas:
@yegor256 plz review

0crat:
@vasilevnstas Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (7) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3080](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #736, opened 2026-09-01
#734: divide the earned value as a fraction

### LINKED ISSUE #734: Incorrect CPI and SPI values in EVA chart due to integer division in `index-eva.rb`
## What happens

The `index-eva` judge divides two Factbase properties that are `Integer` in production,
so Ruby performs integer division and truncates the result toward zero before it is
stored:

https://github.com/zerocracy/pages-action/blob/master/judges/index-eva/index-eva.rb#L22-L25

```ruby
).each do |f|
  f.n_cpi = f.ev / f.ac
  f.n_spi = f.ev / f.pv
end
```

`vitals.xsl` plots those stored values in the EVA chart:

https://github.com/zerocracy/pages-action/blob/master/xsl/vitals.xsl#L247

```xml
<xsl:with-param name="colors" select="'n_spi:darkred,n_cpi:darkblue'"/>
```

while the paragraph printed directly above that chart computes the same two indexes in
XPath, where `div` is floating point:

https://github.com/zerocracy/pages-action/blob/master/xsl/eva-summary.xsl#L14-L17

```xml
<xsl:text>, CPI: </xsl:text>
<xsl:copy-of select="z:index(ev div ac)"/>
<xsl:text>, SPI: </xsl:text>
<xsl:copy-of select="z:index(ev div pv)"/>
```

The result is a page that disagrees with itself. On the currently published
[objectionary vitals page](https://www.eolang.org/zerocracy/objectionary-vitals.html)
the text reads:

```
AC: 36422, EV: 301659, PV: 66822, CPI: +8.28, SPI: +4.51.
```

and the chart rendered underneath it, from the same two facts, carries:

```js
datasets:[{label:'Cpi',borderColor:'darkblue',data:[0,0,3,3,2,3,1,4,1,2,2,9,8]},
          {label:'Spi',borderColor:'darkred',data:[0,0,0,1,1,0,1,0,0,0,0,0,4]}]
```

SPI is the worse half. Schedule performance is normally a ratio near `1`, so truncation
collapses it: of the 13 `earned-value` facts in the published objectionary factbase, **nine
have a true SPI between 0.48 and 0.76, and all nine are plotted as `0`**. The darkred line
is flat on zero for most of the chart, which reads as "no schedule progress at all" rather
than "slightly behind schedule".

This is not a fixture artifact — every `earned-value` fact in both published factbases has
integer `ev`, `ac` and `pv`. Recomputing the truncation over
[`objectionary.json`](https://raw.githubusercontent.com/objectionary/eo/gh-pages/zerocracy/objectionary.json)
reproduces the two arrays above point for point, all 13 of 13.

This is separate from #685, which was about `eva-summary.xsl` failing when `ac` or `pv` is
zero; the facts here have non-zero divisors and nothing fails — the stored numbers are
simply wrong.

The judge's own test does not catch it because `add-indexes.yml` supplies floats
(`ev: 4342.42`, `pv: 6540.56`) where production supplies integers, and its assertions only
check that the properties exist (`/fb/f[n_spi]`, `/fb/f[n_cpi]`), never what they equal.

## Steps to reproduce

Values below are copied verbatim from the published factbase.

```ruby
require 'factbase'

real = [
  [6308, 14810, 8354],       # 2026-04-06
  [14404, 15093, 29891],     # 2026-04-16
  [39005, 21567, 63951],     # 2026-06-25
  [301659, 36422, 66822]     # 2026-08-05, the pair printed in the paragraph
]

fb = Factbase.new
real.each do |ev, ac, pv|
  f = fb.insert
  f.what = 'earned-value'
  f.ev = ev
  f.ac = ac
  f.pv = pv
end

fb.query('(eq what "earned-value")').each do |f|
  f.n_cpi = f.ev / f.ac   # the two lines of index-eva.rb, verbatim
  f.n_spi = f.ev / f.pv
  puts format(
    'ev=%-8s ac=%-7s pv=%-8s | stored cpi=%-3s spi=%-3s | true %.2f / %.2f',
    f.ev, f.ac, f.pv, f.n_cpi, f.n_spi, f.ev.to_f / f.ac, f.ev.to_f / f.pv
  )
end
```

## Observed

```
ev=6308     ac=14810   pv=8354     | stored cpi=0   spi=0   | true 0.43 / 0.76
ev=14404    ac=15093   pv=29891    | stored cpi=0   spi=0   | true 0.95 / 0.48
ev=39005    ac=21567   pv=63951    | stored cpi=1   spi=0   | true 1.81 / 0.61
ev=301659   ac=36422   pv=66822    | stored cpi=8   spi=4   | true 8.28 / 4.51
```

## Expected behavior

`n_cpi` and `n_spi` should hold the same values the paragraph above the chart shows, so
that a CPI of `8.28` is plotted as `8.28` and an SPI of `0.61` is plotted as `0.61`
instead of `0`. The division should not depend on whether the incoming `ev`, `ac` and
`pv` happen to be integers or floats.

morphqdd:
`f.ev / f.ac` and `f.ev / f.pv` are integer divisions when the three
properties are integers, which they are in production, so a CPI of 1.25 was
stored as 1 and an SPI of 0.83 as 0. The chart drawn from those values
disagreed with the paragraph above it, which computes the same two indexes in
XPath, where division is floating point.

The dividend is a float now, so the quotient keeps its fraction.

The pack gained a fifth input with integer `ac`, `ev` and `pv`, and two
asserts on the indexes it produces; both fail on master as it is.

Closes #734

### DIFF
--- a/judges/index-eva/add-indexes.yml
+++ b/judges/index-eva/add-indexes.yml
@@ -31,7 +31,16 @@ input:
     ac: 433
     ev: 437
     pv: 0
+  -
+    start: 2024-04-17T22:22:22.8492Z
+    when: 2024-05-17T22:22:22.8492Z
+    what: earned-value
+    ac: 444
+    ev: 555
+    pv: 666
 expected:
-  - /fb[count(f)=4]
+  - /fb[count(f)=5]
   - /fb/f[n_spi]
   - /fb/f[n_cpi]
+  - /fb/f[ac=444 and n_cpi > 1.2 and n_cpi < 1.3]
+  - /fb/f[ac=444 and n_spi > 0.8 and n_spi < 0.9]
--- a/judges/index-eva/index-eva.rb
+++ b/judges/index-eva/index-eva.rb
@@ -20,6 +20,6 @@ Fbe.fb.query(
     (absent n_spi))
   "
 ).each do |f|
-  f.n_cpi = f.ev / f.ac
-  f.n_spi = f.ev / f.pv
+  f.n_cpi = f.ev.to_f / f.ac
+  f.n_spi = f.ev.to_f / f.pv
 end


yegor256:
Thanks for tracking this down — using `f.ev.to_f` so the division stays floating point (matching what `eva-summary.xsl` already computes) looks like the right fix, and the new `add-indexes.yml` case pins down the CPI/SPI values from #734 directly.

However, the `xcop` check is currently failing on this PR (https://github.com/zerocracy/pages-action/actions/runs/33545653094/job/99982363125), so I can't merge as-is. Could you take a look at what it's flagging (likely XML formatting in `add-indexes.yml`) and push a fix? I'm requesting @yegor256 to review once CI is green.

---

morphqdd:

PR #747, opened 2026-09-03
#743: use the supplied logo as the favicon

### LINKED ISSUE #743: `adless` parameter ignored causing the Zerocracy favicon to display instead of custom logos
## What happens

https://github.com/zerocracy/pages-action/blob/master/xsl/vitals.xsl#L183-L185

```xml
<xsl:if test="$logo != ''">
  <link rel="icon" href="https://www.zerocracy.com/svg/logo.svg" type="image/svg"/>
</xsl:if>
```

The condition tests `$logo` and the `href` ignores it. `action.yml:30-32` documents `adless` as
"Hide all Zerocracy links, banners, and logos" and `entry.sh:194-198` prints that the output will
have no mention of Zerocracy. Every other Zerocracy reference in the file sits behind
`<xsl:if test="$adless = 'false'">`; this one does not.

```
$ java -jar target/saxon.jar -s:/tmp/zcv/probe2.xml -xsl:xsl/vitals.xsl -o:/tmp/zcv/out5.html \
    today=2024-07-05T00:00:00Z ... logo=https://example.com/mylogo.svg ... adless=true ...
$ grep -oE 'zerocracy[a-z.]*' /tmp/zcv/out5.html | sort | uniq -c
      1 zerocracy.com
$ grep -n -A2 'rel="icon"' /tmp/zcv/out5.html
13:      <link rel="icon"
14-            href="https://www.zerocracy.com/svg/logo.svg"
15-            type="image/svg"/>
```

One `zerocracy.com` on an `adless=true` page, and it is this favicon: a third-party request and a
visible brand leak. A project that sets a custom logo without `adless` also gets the Zerocracy icon
instead of its own.

## What should happen

Use the supplied logo, `<link rel="icon" href="{$logo}" type="image/svg"/>`, and fall back to the
Zerocracy icon only when `$adless = 'false'`, the same gate `entry.sh:180-184` already uses to pick
the default logo.

morphqdd:
The condition tested the `logo` parameter while the `href` ignored it and always
pointed at the Zerocracy icon, so an adless page still made a third-party request
for it and a project with its own logo got the wrong icon.

`entry.sh` already picks the Zerocracy logo as the default only when `adless` is
false, so using the parameter keeps that fallback.

Closes #743

### DIFF
--- a/entries/renders-adless-vitals.sh
+++ b/entries/renders-adless-vitals.sh
@@ -16,12 +16,14 @@ env "GITHUB_WORKSPACE=$(pwd)" \
   'INPUT_FACTBASE=test.fb' \
   'INPUT_OUTPUT=output' \
   'INPUT_VERBOSE=false' \
-  'INPUT_LOGO=' \
+  'INPUT_LOGO=https://example.com/mylogo.svg' \
   'INPUT_ADLESS=true' \
   'INPUT_GITHUB-TOKEN=THETOKEN' \
   'LATEST_VERSION=0.0.0' \
   "${SELF}/entry.sh" 2>&1 | tee log.txt
 
 grep "The output will have no mention of Zerocracy" 'log.txt'
+
+grep 'rel="icon"' -A 1 'output/test-vitals.html' | grep 'https://example.com/mylogo.svg'
 
 grep -v zerocracy 'output/test-vitals.html'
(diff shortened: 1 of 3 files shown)


yegor256:
Making the favicon `href` actually respect the `logo` parameter is a clear, correct fix. I can't merge as-is though: the `xcop` check is failing on the head commit (`Invalid XML formatting in xsl/assessment.xsl`), a file this PR doesn't touch. That looks like a pre-existing formatting issue rather than something this change introduced, but I'd rather have a maintainer confirm than override CI. Reassigning for manual review.

---

morphqdd:

PR #748, opened 2026-09-03
#744: assert that the adless page names nobody

### LINKED ISSUE #744: `renders-adless-vitals.sh` inaccurately tests for `zerocracy` references with `grep -v` leading to false positives
## What happens

https://github.com/zerocracy/pages-action/blob/master/entries/renders-adless-vitals.sh#L27

```bash
grep -v zerocracy 'output/test-vitals.html'
```

The script runs under `set -e` and leans on this line to prove the adless page has no Zerocracy
references. `grep -v` inverts which lines are printed, not the exit status: it exits 0 as soon as
one line does not match, which on an HTML page is always.

```
$ printf 'zerocracy\nfoo\n' > /tmp/zcv/g.txt
$ grep -v zerocracy /tmp/zcv/g.txt; echo "exit=$?"
foo
exit=0
```

Exit 0 on a file that plainly contains the word. The script also passes an empty `INPUT_LOGO`, so
it never reaches the `<xsl:if test="$logo != ''">` branch at all.

## What should happen

Assert absence: `! grep -q zerocracy 'output/test-vitals.html'`. And run the script once with a
non-empty `INPUT_LOGO` so the favicon branch is covered.

morphqdd:
`grep -v` inverts which lines are printed, not the exit status, so the check
exited 0 on any HTML page and proved nothing. It asserts absence now.

With a real assertion the page turns out not to be clean: the `description` and
`og:description` meta tags say "supervised by Zerocracy" whatever `adless` is.
That wording is now behind the same gate as every other mention, so the adless
page names nobody, and a test renders it and checks that.

The favicon branch needs a non-empty logo, which leaks the Zerocracy icon until
the fix in #743 lands, so that part of the coverage goes with it.

Closes #744

### DIFF
--- a/entries/renders-adless-vitals.sh
+++ b/entries/renders-adless-vitals.sh
@@ -24,4 +24,4 @@ env "GITHUB_WORKSPACE=$(pwd)" \
 
 grep "The output will have no mention of Zerocracy" 'log.txt'
 
-grep -v zerocracy 'output/test-vitals.html'
+! grep -qi zerocracy 'output/test-vitals.html'
--- a/test/pages/test_vitals.rb
+++ b/test/pages/test_vitals.rb
@@ -142,9 +142,13 @@ class TestVitals < Minitest::Test
     refute_match(/zerocracy/i, html[%r{<meta name="description".*?/>}m].to_s, html)
   end
 
+  def test_adless_page_does_not_name_zerocracy
+    refute_match(/zerocracy/i, generate_vitals_html(adless: 'true', logo: ''))
+  end
+
   private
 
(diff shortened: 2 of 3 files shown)


yegor256:
Turning `grep -v` into a real assertion, and closing the "supervised by Zerocracy" leak on the adless page, is good work. I can't merge as-is though: the `xcop` check is failing on the head commit (`Invalid XML formatting in xsl/assessment.xsl`), a file this PR doesn't touch. That looks like a pre-existing formatting issue rather than something this change introduced, but I'd rather have a maintainer confirm than override CI. Reassigning for manual review.

---

morphqdd:

PR #749, opened 2026-09-03
#742: sort the leaderboard by the group total

### LINKED ISSUE #742: Leaderboard displays incorrect order due to sort key reading only the first fact
## What happens

https://github.com/zerocracy/pages-action/blob/master/xsl/awards.xsl#L241-L242

```xml
<xsl:for-each-group select="$facts" group-by="who_name">
  <xsl:sort select="sum(award)" data-type="number" order="descending"/>
```

Inside `for-each-group` a sort key is evaluated with the context item set to the first item of the
group, not the group, so `sum(award)` sums one fact's awards — that programmer's first one. The
"sort by total score" asked for in #30 does not happen.

Isolated probe of the semantics:

```
$ java -jar target/saxon.jar -s:in.xml -xsl:sortprobe.xsl
bob sortkey_used=3 real_total=3
alice sortkey_used=1 real_total=4
```

End to end, alice with four awards of 1 and bob with one of 3:

```
$ grep -oE 'data-value="[0-9]+"|@(alice|bob)</a>' /tmp/zcv/out2.html
@bob</a>            data-value="0" x3  data-value="3"  data-value="3"
@alice</a>          data-value="0" x3  data-value="4"  data-value="4"
```

Bob, with 3, is rendered above alice, with 4, under a descending sort.

Scope note: `awards.xsl:165-172` sets up tablesorter with `sortList: [[2 + weeks, 1]]`, which
re-sorts by `Run` once the script runs, so a browser masks it. The static HTML, which is what a
crawler, a saved page or a reader without JavaScript sees, is ordered wrong, and the XSLT sort is
not doing what it says.

## What should happen

`<xsl:sort select="sum(current-group()/award)" data-type="number" order="descending"/>`. While
there, `awards.xsl:245` is true by construction of the group and can go.

morphqdd:
Inside `for-each-group` a sort key is evaluated with the context item set to the
first item of the group, so `sum(award)` summed one fact, that programmer's first
award. The static page came out in the wrong order; the browser only hides it
because tablesorter re-sorts after load.

The key is `sum(current-group()/award)` now. The `count(...) > 0` guard next to it
is true by construction of the group, so it is gone.

Closes #742

### DIFF
--- a/test/pages/test_awards.rb
+++ b/test/pages/test_awards.rb
@@ -224,4 +224,27 @@ class TestAwards < Minitest::Test
     )
     assert_empty(xml.xpath("//a[starts-with(@href, 'javascript:')]"), xml)
   end
+
+  def test_leaderboard_sorts_by_the_total
+    now = Time.now.utc
+    xml = xslt(
+      '<xsl:apply-templates select="/" mode="awards"/>',
+      "
+      <fb>
+        <f><is_human>1</is_human><who>1</who><who_name>alice</who_name><award>1</award>
+          <when>#{(now - (60 * 60)).iso8601}</when></f>
+        <f><is_human>1</is_human><who>1</who><who_name>alice</who_name><award>1</award>
+          <when>#{(now - (60 * 60)).iso8601}</when></f>
+        <f><is_human>1</is_human><who>1</who><who_name>alice</who_name><award>1</award>
+          <when>#{(now - (60 * 60)).iso8601}</when></f>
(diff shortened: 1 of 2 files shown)


yegor256:
Good catch on the `for-each-group` sort-key context bug. I can't merge as-is though: the `xcop` check is failing on the head commit (`Invalid XML formatting in xsl/assessment.xsl`), a file this PR doesn't touch. That looks like a pre-existing formatting issue rather than something this change introduced, but I'd rather have a maintainer confirm than override CI. Reassigning for manual review.

---

morphqdd:

PR #750, opened 2026-09-03
#741: pick the earned-value fact that happened last

### LINKED ISSUE #741: EVA summary inaccurately displays latest fact due to document-order sort over string-sorted `_id`
## What happens

https://github.com/zerocracy/pages-action/blob/master/xsl/vitals.xsl#L249

```xml
<xsl:apply-templates select="/fb/f[what='earned-value'][last()]"/>
```

`[last()]` is last in document order, not latest by `when`. Document order comes from
`Factbase::Flatten#it` (`factbase-0.19.16/lib/factbase/flatten.rb:24`), which does
`sort_by { |m| Array(m[@sorter]).map(&:to_s) }` — a string sort of `_id`. So `_id=10` sorts before
`_id=9`, and the fact chosen as "latest" is whichever sorts last as a string.

Two facts, `_id=10` at 2024-07-04 with `ev=90`, and `_id=9` at 2024-06-01 with `ev=10`:

```
$ grep -E "_id|<when" /tmp/zcv/probe7.xml
    <_id t="I">10</_id>
    <when t="T">2024-07-04T10:00:00Z</when>
    <_id t="I">9</_id>
    <when t="T">2024-06-01T10:00:00Z</when>

$ grep -o "AC: [^<]*<span[^>]*>[^<]*</span>[^<]*<span[^>]*>[^<]*</span>." /tmp/zcv/out7.html
AC: 100, EV: 10, PV: 100, CPI: <span class="darkgreen">+0.10</span>, SPI: <span class="darkgreen">+0.10</span>.
```

The page reports June's CPI and SPI while July's fact is the current one.

## What should happen

Select by time rather than by position, sorting the facts on `xs:dateTime(when)` and taking the
maximum. `awards.xsl:43` picks the last reconciliation the same way and has the same problem.

morphqdd:
`[last()]` is last in document order, and document order comes from a string sort
of `_id`, so `_id=10` sorts before `_id=9` and the page reported an older EVA fact
as the current one. The reconciliation lookup in `awards.xsl` had the same
problem.

Both go through one `z:latest` function now, which picks by `when`.

Closes #741


yegor256:
Good find on the string-sort-vs-`when` ordering bug, and consolidating into one `z:latest` function is a nice cleanup, but I can't merge as-is: the `xcop` check is failing on the head commit (`Invalid XML formatting in xsl/assessment.xsl`), a file this PR doesn't touch. That looks like a pre-existing formatting issue rather than something this change introduced, but I'd rather have a maintainer confirm than override CI. Reassigning for manual review.

---

morphqdd:

PR #751, opened 2026-09-03
#739: count only human awards on the badge and in the meta

### LINKED ISSUE #739: Inconsistency in counting awards between the badge, meta, and table
## What happens

The awards table counts human awards only, and two other places on the same page do not.

```
awards.xsl:11   $fb/f[award and xs:dateTime(when) > $since and is_human = 1]
badge.xsl:11    .//f[xs:dateTime(when) > $since and award]
vitals.xsl:146  $fb/f[xs:dateTime(when) > (...P256D) and award]
```

So the badge average, the `description` meta and the `og:description` include bot awards, while
the table right below them excludes them. `vitals.xsl:156` counts bots as contributors too.

One human award of 10 and one bot award of 1000, both on 2024-07-02:

```
$ grep -o "supervised by Zerocracy[^\"]*" /tmp/zcv/out8.html | head -1
supervised by Zerocracy: +505.0 average points per task, 1010 total points earned, 2 contributors.
$ grep -oE ">[+-][0-9.]+</text>" /tmp/zcv/badge8.svg | head -1
>+505</text>
$ grep -o "There are [0-9]* awarding facts in the XML, while [0-9]* of them count" /tmp/zcv/out8.html
There are 2 awarding facts in the XML, while 1 of them count
$ grep -o 'data-value="[0-9]*"' /tmp/zcv/out8.html | tail -1
data-value="10"
```

Badge `+505`, meta `+505.0 average, 1010 total, 2 contributors`, table `Run` 10 and "1 of them
count". The README says the badge shows the average reward and that a higher number means better
discipline, so a bot can move it at will.

## What should happen

All three should use the same population. Add `and is_human = 1` to `badge.xsl:11` and
`vitals.xsl:146`, or hoist the predicate into one shared place so the three cannot drift again.

morphqdd:
The awards table counts human awards only, while the badge and the `description`
and `og:description` meta tags counted bots too, so the three numbers on one page
disagreed and a bot could move the badge at will.

All three use the same population now.

Closes #739

### DIFF
--- a/entries/renders-vitals-description.sh
+++ b/entries/renders-vitals-description.sh
@@ -16,16 +16,19 @@ bundle exec judges eval test.fb "
   f.award = 10
   f.who = 12345
   f.who_name = 'user1'
+  f.is_human = 1
   f = \$fb.insert
   f.when = '2024-06-26T00:00:00Z'
   f.award = 20
   f.who = 12345
   f.who_name = 'user1'
+  f.is_human = 1
   f = \$fb.insert
   f.when = '2023-10-22T00:00:00Z'
   f.award = 100
   f.who = 67890
   f.who_name = 'user2'
+  f.is_human = 1
 " > /dev/null
 
 env "GITHUB_WORKSPACE=$(pwd)" \
--- a/test/pages/test_badge.rb
+++ b/test/pages/test_badge.rb
@@ -33,6 +33,18 @@ class TestBadge < Minitest::Test
     assert_includes(texts, '+0', xml)
   end
 
+  def test_ignores_bot_awards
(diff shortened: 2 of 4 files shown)


yegor256:
Making the badge and meta tags agree with the awards table on bot filtering is a good fix, but I can't merge as-is: the `xcop` check is failing on the head commit (`Invalid XML formatting in xsl/assessment.xsl`), a file this PR doesn't touch. That looks like a pre-existing formatting issue rather than something this change introduced, but I'd rather have a maintainer confirm than override CI. Reassigning for manual review.

---

morphqdd:

PR #752, opened 2026-09-03
#740: write null for a data point that is absent

### LINKED ISSUE #740: `qo-section.xsl` emits an array hole instead of `null` when data point is missing
## What happens

https://github.com/zerocracy/pages-action/blob/master/xsl/qo-section.xsl#L115-L123

```xml
<xsl:variable name="cell" select="*[name()=$n]/text()"/>
<xsl:choose>
  <xsl:when test="$cell = ''"><xsl:text>null</xsl:text></xsl:when>
  <xsl:otherwise><xsl:value-of select="$cell"/></xsl:otherwise>
</xsl:choose>
```

When the fact does not carry the property, `$cell` is the empty sequence, and `() = ''` is false in
XPath, so the `null` branch is dead and `value-of` on an empty sequence writes nothing. The
Chart.js array gets a hole instead of a value.

This is the expected output of the judge that feeds it: `normalize-metrics.rb:35` skips a metric
whose baseline is zero (`next if start[prop].zero?`), the fix accepted in #593.

Three weekly facts with the middle one missing `n_average_issue_lifetime`:

```
$ grep -o "qo_render('quality-of-service'.*" /tmp/zcv/out3.html
qo_render('quality-of-service',{labels:['6/10','6/17','6/24'],...,data:[0.2,,0.5]},...)
```

Worse when the last fact is the one missing it:

```
$ grep -o "qo_render('quality-of-service'.*" /tmp/zcv/out4.html
... data:[0.2,,]}, ...
$ node -e "console.log([0.2,0.3,].length)"
2
```

Three labels, an array of length two, so the series is silently misaligned against the x-axis.

## What should happen

Test for absence rather than for an empty string, `<xsl:when test="empty($cell)">`. Chart.js
renders `null` as a documented gap.

morphqdd:
When the fact does not carry the property, `$cell` is the empty sequence, and
`() = ''` is false in XPath, so the `null` branch never ran and `value-of` wrote
nothing. The Chart.js array got a hole, and a hole in the last position shortens
the array, which shifts the whole series against its labels.

The judge that feeds this skips a metric whose baseline is zero, so the absent
property is expected input.

Closes #740

### DIFF
--- a/test/pages/test_qo_section.rb
+++ b/test/pages/test_qo_section.rb
@@ -94,4 +94,32 @@ class TestQoSection < Minitest::Test
       end
     end
   end
+
+  def test_missing_data_point_is_null
+    xml = xslt(
+      "<r><xsl:call-template name='qo-section'>" \
+      "<xsl:with-param name='what' select=\"'quality-of-service'\"/>" \
+      "<xsl:with-param name='title' select=\"'QoS'\"/>" \
+      '</xsl:call-template></r>',
+      '<fb>
+        <f>
+          <when>2024-06-23T22:22:22Z</when>
+          <what>quality-of-service</what>
+          <n_composite>0.2</n_composite>
+        </f>
+        <f>
+          <when>2024-06-30T22:22:22Z</when>
+          <what>quality-of-service</what>
+        </f>
+        <f>
+          <when>2024-07-03T22:22:22Z</when>
+          <what>quality-of-service</what>
+          <n_composite>0.5</n_composite>
+        </f>
+      </fb>',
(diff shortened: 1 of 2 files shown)


yegor256:
Good catch on the empty-sequence-vs-empty-string XPath gap, but I can't merge as-is: the `xcop` check is failing on the head commit (`Invalid XML formatting in xsl/assessment.xsl`), a file this PR doesn't touch. That looks like a pre-existing formatting issue rather than something this change introduced, but I'd rather have a maintainer confirm than override CI. Reassigning for manual review.

---

morphqdd:

PR #753, opened 2026-09-03
#745: assert what the index judge computes

### LINKED ISSUE #745: `index-eva` pack incorrectly asserts existence of `n_cpi` and `n_spi`, doesn't validate their values
## What happens

https://github.com/zerocracy/pages-action/blob/master/judges/index-eva/add-indexes.yml#L34-L37

```yaml
expected:
  - /fb[count(f)=4]
  - /fb/f[n_spi]
  - /fb/f[n_cpi]
```

Both are existential over the whole factbase, so one fact carrying the property anywhere satisfies
them. The pack checks that the judge wrote something, never what.

Baseline:

```
$ bundle exec judges test --judge index-eva judges
  index-eva/add-indexes    0.051   OK
```

Replace both computations with a value that is obviously not a CPI or an SPI:

```
$ sed -i 's|f.n_cpi = f.ev / f.ac|f.n_cpi = -12345.0|; s|f.n_spi = f.ev / f.pv|f.n_spi = -12345.0|' judges/index-eva/index-eva.rb
$ bundle exec judges test --judge index-eva judges
  index-eva/add-indexes    0.050   OK
```

Still green. This is why the integer division in #734 shipped: `f.ev / f.ac` on integers passes
this pack exactly as well as the correct division does.

## What should happen

Assert the values. The sibling `normalize-metrics` packs already do, for instance
`/fb/f[composite = 120 and n_composite = 0.2]`. For the fact with `ac: 433, ev: 4342.42, pv: 6540.56`
the pack should pin `n_cpi` and `n_spi` to their expected floats, and should also assert that the
two skipped facts, one with no `ac` and one with `pv = 0`, got no `n_cpi` or `n_spi`, since nothing
checks the guard clauses today.

morphqdd:
`/fb/f[n_spi]` and `/fb/f[n_cpi]` are existential over the whole factbase, so one
fact carrying the property anywhere satisfied them. The pack checked that the
judge wrote something, never what, and never that it left the other facts alone.

It now asserts how many facts got the indexes, the values for the two facts that
qualify, and that the fact without `ac` and the one with a zero `pv` stay
untouched. Swapping `ac` for `pv` in the judge makes it fail.

Closes #745

### DIFF
--- a/judges/index-eva/add-indexes.yml
+++ b/judges/index-eva/add-indexes.yml
@@ -35,7 +35,9 @@ expected:
   - /fb[count(f)=4]
   - /fb[count(f[n_cpi])=2]
   - /fb[count(f[n_spi])=2]
-  - /fb/f[ac='433' and ev='4342.42' and starts-with(n_cpi,'10.02') and starts-with(n_spi,'0.66')]
+  - >
+    /fb/f[ac='433' and ev='4342.42' and starts-with(n_cpi,'10.02')
+    and starts-with(n_spi,'0.66')]
   - /fb/f[ev='0.0' and n_cpi='0.0' and n_spi='0.0']
   - /fb/f[not(ac) and not(n_cpi) and not(n_spi)]
   - /fb/f[pv='0' and not(n_cpi) and not(n_spi)]
--- a/judges/index-eva/add-indexes.yml
+++ b/judges/index-eva/add-indexes.yml
@@ -33,5 +33,9 @@ input:
     pv: 0
 expected:
   - /fb[count(f)=4]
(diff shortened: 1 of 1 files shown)


yegor256:
This is a solid, well-targeted test fix, but I can't merge as-is: the `xcop` check is failing on the head commit (`Invalid XML formatting in xsl/assessment.xsl`), a file this PR doesn't touch. That looks like a pre-existing formatting issue rather than something this change introduced, but I'd rather have a maintainer confirm than override CI. Reassigning for manual review.

---

morphqdd:

PR #754, opened 2026-09-03
#738: cover the same interval in the facts and the columns

### LINKED ISSUE #738: Week columns and running-balance window are covering different intervals, causing inconsistencies
## What happens

The fact window and the week columns cover different intervals.

https://github.com/zerocracy/pages-action/blob/master/xsl/awards.xsl#L9-L11

`$since` is `today - P{days}D`, a rolling window ending today, while each column is anchored to a
Monday by `z:monday($week)`. Unless today is a Monday the first column starts after `$since`, so an
award in that gap is counted in `$facts` and in the `Run` total (`:284`, `:330`) and in no column.
With the default 28 days the gap is `weekday(today) - 1` days, so it exists six days out of seven.
`z:in-week` at `:35` is also doubly strict, `> $monday and < $monday + P7D`, so an award at exactly
Monday 00:00:00 lands in no week either.

Factbase built through the real pipeline, one award of 100 at `2024-06-08` (inside `$since` =
`2024-06-07`), one of 1 at `2024-07-02`, one of 5 at `2024-07-03`, today `2024-07-05`, a Friday:

```
$ java -jar target/saxon.jar -s:/tmp/zcv/probe.xml -xsl:xsl/vitals.xsl -o:/tmp/zcv/out.html today=2024-07-05T00:00:00Z ...
```

```html
<th title="Week #24 in 2024, starting on Monday 2024-06-10Z">w24</th> ... w25 ... w26 ... w27
<td data-value="0">—</td><td data-value="0">—</td><td data-value="0">—</td><td data-value="6">+6</td>
<td data-value="106">+106</td>  <!-- Run -->
```

0 + 0 + 0 + 6 is 6, and `Run` says 106. The first column starts on Monday 2024-06-10, three days
after `$since`. The contributor's detail sub-row for the 100-point award is blank in every week
cell.

## What should happen

The facts and the columns should cover the same interval: derive `$since` from the first column's
Monday, or add the leading partial week as its own column. And `z:in-week` should be half-open,
`>= $monday and < $monday + P7D`, so no timestamp falls between two weeks.

morphqdd:
`$since` was a rolling window ending today while each column is anchored to a
Monday, so unless today is a Monday the first column started after `$since` and
an award in that gap was counted in `Run` and in no column. `z:in-week` was also
strict on both ends, so an award at exactly Monday 00:00:00 fell between two
weeks.

`$since` is now the Monday of the first column, and a week is half-open, so the
columns cover exactly what `Run` sums.

Closes #738

### DIFF
--- a/test/pages/test_awards.rb
+++ b/test/pages/test_awards.rb
@@ -224,4 +224,24 @@ class TestAwards < Minitest::Test
     )
     assert_empty(xml.xpath("//a[starts-with(@href, 'javascript:')]"), xml)
   end
+
+  def test_week_columns_add_up_to_the_run
+    xml = xslt(
+      '<xsl:apply-templates select="/" mode="awards"/>',
+      '
+      <fb>
+        <f><is_human>1</is_human><who>1</who><who_name>alice</who_name><award>100</award>
+          <when>2024-06-08T10:00:00Z</when></f>
+        <f><is_human>1</is_human><who>1</who><who_name>alice</who_name><award>7</award>
+          <when>2024-07-01T00:00:00Z</when></f>
+        <f><is_human>1</is_human><who>1</who><who_name>alice</who_name><award>5</award>
+          <when>2024-07-03T10:00:00Z</when></f>
+      </fb>
+      ',
+      'today' => '2024-07-05T00:00:00Z'
+    )
+    values = xml.xpath('//tbody/tr[1]/td[@data-value]').map { |t| t['data-value'].to_i }
+    assert_equal(values.last, values[0..-2].sum, xml)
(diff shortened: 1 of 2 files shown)


yegor256:
The fix itself reads well and is well scoped, but I can't merge as-is: the `xcop` check is failing on the head commit (`Invalid XML formatting in xsl/assessment.xsl`), a file this PR doesn't touch. That looks like a pre-existing formatting issue rather than something this change introduced, but I'd rather have a maintainer confirm than override CI. Reassigning for manual review.

---

morphqdd:

PR #776, opened 2026-09-07
#772: name only the facts that need a name

### LINKED ISSUE #772: `copy-who-names` incorrectly duplicates `who_name` in `who-has-name` facts
## What happens

`judges/copy-who-names/copy-who-names.rb:16` finds the facts to annotate:

```ruby
Fbe.fb.query('(and (exists who) (not (exists who_name)))').each do |f|
```

There is no `what` restriction, and the `who-has-name` facts themselves have a `who`, so the
name registry matches its own lookup table and gets annotated with a copy of the name it
already holds:

```
#0: {"what"=>["who-has-name"], "who"=>[444], "name"=>["travolta"], "where"=>["github"],
     "who_name"=>["travolta"]}
#1: {"what"=>["something"], "who"=>[444], "where"=>["github"], "who_name"=>["travolta"]}
```

Fact #0 is the registry entry, and it did not need a `who_name`. The duplicate property is
carried in the factbase and printed in the YAML, XML and JSON dumps that `entry.sh:105-113`
produces.

## What should happen

The query should exclude the registry facts, for example with
`(not (eq what "who-has-name"))`, so only the facts that need a name get one.

morphqdd:
The query had no `what` restriction, and a `who-has-name` fact has a `who`, so the name
registry matched its own lookup table and got a copy of the name it already held. The
duplicate property then travelled into every YAML, XML and JSON dump the entry script writes.

The pack now checks both sides: the `something` fact gets the name, the `who-has-name` fact
does not.

Closes #772

### DIFF
--- a/judges/copy-who-names/copy-who-names.rb
+++ b/judges/copy-who-names/copy-who-names.rb
@@ -13,7 +13,12 @@ names =
       (exists name))'
   ).each.to_a.to_h { |f| [f.who, f.name] }
 
-Fbe.fb.query('(and (exists who) (not (exists who_name)))').each do |f|
+Fbe.fb.query(
+  '(and
+    (not (eq what "who-has-name"))
+    (exists who)
+    (not (exists who_name)))'
+).each do |f|
   n = names[f.who]
   next if n.nil?
   f.who_name = n
--- a/judges/copy-who-names/name-exists.yml
+++ b/judges/copy-who-names/name-exists.yml
@@ -13,4 +13,5 @@ input:
     where: github
 expected:
   - /fb[count(f)=2]
-  - /fb/f[who_name='travolta']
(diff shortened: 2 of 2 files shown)


yegor256:
Merged, thanks! Excluding `who-has-name` from the query itself is cleaner than filtering afterwards, and the pack now pins both halves of the expectation.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+8004](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #683, opened 2026-06-30
badge.xsl displays "-0" when average score is exactly zero


sdghsdkjlas27-dotcom:
badge.xsl displays "-0" when average score is exactly zero

DESCRIPTION:
The badge XSLT template uses strict greater-than (`> 0`) for both sign selection and color selection. When the average is exactly `0.0` (which occurs when the fact count is zero or when `format-number` rounds a tiny negative to zero), the sign logic falls through to the otherwise branch and prepends `-`, while `abs(0)` produces `0`. The badge renders the text "-0".

REPRODUCTION:
`xsl/badge.xsl`, lines 71-78:
```xml
<xsl:when test="$avg &gt; 0">
  <xsl:text>+</xsl:text>
</xsl:when>
<xsl:otherwise>
  <xsl:text>-</xsl:text>
</xsl:otherwise>
```
When `$avg` is exactly `"0.0"` (from `format-number()`), `$avg > 0` is false, so the otherwise-branch outputs `-`. Then `abs($avg)` outputs `0`. Result: "-0".

Compare with `vitals.xsl` line 56 which correctly uses `>= 0`:
```xml
<xsl:when test="$i &gt;= 0">
  <xsl:text>darkgreen</xsl:text>
```

IMPACT:
Medium — incorrect badge display. Zero should not show a negative sign.

RELATED FILES:
- `xsl/badge.xsl` (lines 71-78)
- `xsl/vitals.xsl` (line 56 — correct `>= 0` pattern)

PR #777, opened 2026-09-07
#766: print the average with the decimal it was formatted with

### LINKED ISSUE #766: Whole average value loses decimal in badge due to `abs()` re-parsing formatted string
## What happens

`xsl/badge.xsl:22` builds the average as a formatted string with one decimal:

```xml
<xsl:value-of select="format-number($sum div $count, '0.0')"/>
```

then line 84 prints it back through `abs()`:

```xml
<xsl:value-of select="abs($avg)"/>
```

`abs()` takes a number, so the string is cast back to a double and the formatting is undone.
A whole average loses its decimal. Two awards of 4:

```
$ ... -xsl:xsl/badge.xsl today=2026-09-06T00:00:00Z | grep text-anchor
avg
+4
+4
```

The empty-factbase branch (line 17) writes `0.0` and reaches the same `abs()`, so it prints
`+0`, while the `description` meta on the vitals page keeps saying `+0.0 average points per
task`. So the badge is `+4` for an average of 4, `+4.5` for an average of 4.5, and disagrees
with the page next to it.

## What should happen

The badge should show one decimal always, the way `format-number` was asked to. Formatting the
absolute value once (`format-number(abs($sum div $count), '0.0')`) keeps the sign logic and the
`> 999` clamp working without parsing the string back into a number.

morphqdd:
`$avg` is built with `format-number(..., '0.0')`, and then `abs($avg)` cast the string back to a
double, which undid the formatting. A whole average printed as `+4`, while a fractional one kept
its decimal and the `description` meta on the vitals page said `+4.0`.

Formatting the absolute value keeps the sign logic and the clamp untouched, and the clamp text
gains the same decimal so all three branches agree.

One existing assertion moves from `+0` to `+0.0`. #683 was about the badge showing `-0`, and the
badge still shows no minus for a zero average; the `+0` in that assertion was whatever the double
happened to print at the time, not the point being pinned.

Closes #766

### DIFF
--- a/test/pages/test_badge.rb
+++ b/test/pages/test_badge.rb
@@ -30,7 +30,20 @@ class TestBadge < Minitest::Test
     xml = badge_svg('<fb/>')
     texts = xml.xpath("//*[local-name()='text']").map(&:text)
     refute_includes(texts, '-0', "Badge with a zero average must not show a negative sign:\n#{xml}")
-    assert_includes(texts, '+0', xml)
+    assert_includes(texts, '+0.0', xml)
+  end
+
+  def test_shows_one_decimal_for_a_whole_average
+    xml = badge_svg(
+      "
+      <fb>
+        <f><award>4</award><when>2023-12-01T00:00:00Z</when></f>
+        <f><award>4</award><when>2023-12-02T00:00:00Z</when></f>
+      </fb>
+      "
+    )
+    texts = xml.xpath("//*[local-name()='text']").map(&:text)
+    assert_includes(texts, '+4.0', xml)
   end
 
   private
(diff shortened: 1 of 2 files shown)


yegor256:
Merged, thanks! Formatting inside `abs()` instead of re-parsing the formatted string is exactly the right fix.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (19) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+7984](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #783, opened 2026-09-07
#768: leave the raw samples alone

### LINKED ISSUE #768: `normalize-metrics` incorrectly normalizes `some_*` sample arrays using a single sample value
## What happens

`judges/normalize-metrics/normalize-metrics.rb:8-13` decides which properties to normalize:

```ruby
def fits?(name)
  return true if name == 'composite'
  return false unless name.match?(/^[a-z]+_[a-z]+.*$/)
  return false if name.start_with?('n_')
  true
end
```

`quality-of-service` facts carry raw sample arrays named `some_*`, written one value at a time
(`judges-action`'s `quality-of-service/simple-collect.yml` lists `some_issue_lifetime`,
`some_review_time`, `some_backlog_size` and about a dozen more). `fits?('some_issue_lifetime')`
is true, and lines 28 and 36 then read `f[prop].first`, so the judge normalizes one arbitrary
sample of the array:

```
#0: {"when"=>[2024-01-01], "some_issue_lifetime"=>[10.0, 500.0, 20.0],
     "average_issue_lifetime"=>[176.6], "n_some_issue_lifetime"=>[0.0]}
#1: {"when"=>[2024-01-08], "some_issue_lifetime"=>[30.0, 40.0],
     "average_issue_lifetime"=>[35.0], "n_some_issue_lifetime"=>[2.0]}
```

`n_some_issue_lifetime = 2.0` is +200%, and it is 30/10, the first sample of week two over the
first sample of week one. The real average fell from 176.6 to 35. The number carries no
meaning.

`xsl/qo-section.xsl:84` draws a line for every `n_*` property it finds
(`distinct-values($facts/f/*[starts-with(name(), 'n_')]/name())`), so each `some_*` property
becomes an extra series on the quality-of-service chart next to the meaningful `n_average_*`
ones.

## What should happen

`fits?` should reject the raw sample arrays, either by name (`some_`, the way `n_` is already
rejected) or by refusing any property that holds more than one value.

morphqdd:
Quality-of-service facts carry raw sample arrays named `some_*`, written one value at a time by
the collector. `fits?` accepted them, and the judge reads `f[prop].first`, so it normalized one
arbitrary sample against one arbitrary sample of the baseline week. In the added pack that is
`30 / 10`, a reported +200% for a metric whose real average fell from 100 to 200 the other way.

`qo-section.xsl` charts every `n_*` property it finds, so each of those became an extra
meaningless series next to the real `n_average_*` ones.

The name is rejected the way `n_` already is. The new pack checks that no `n_some_*` is written
and that the average next to it still is.

Closes #768

### DIFF
--- a/judges/normalize-metrics/normalize-metrics.rb
+++ b/judges/normalize-metrics/normalize-metrics.rb
@@ -9,6 +9,7 @@ def fits?(name)
   return true if name == 'composite'
   return false unless name.match?(/^[a-z]+_[a-z]+.*$/)
   return false if name.start_with?('n_')
+  return false if name.start_with?('some_')
   true
 end
 
--- a/judges/normalize-metrics/skips-raw-samples.yml
+++ b/judges/normalize-metrics/skips-raw-samples.yml
@@ -1,0 +1,18 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+---
+input:
+  -
+    what: quality-of-service
+    when: 2024-01-01T00:00:00Z
+    some_issue_lifetime: 10
+    average_issue_lifetime: 100
+  -
+    what: quality-of-service
+    when: 2024-01-08T00:00:00Z
+    some_issue_lifetime: 30
+    average_issue_lifetime: 200
+expected:
+  - /fb[count(f)=2]
+  - /fb[count(f[n_some_issue_lifetime])=0]
+  - /fb/f[n_average_issue_lifetime='1.0']


yegor256:
Merged. Thanks — normalizing one arbitrary sample against another was producing numbers pointing the wrong way.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (19) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6182](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #784, opened 2026-09-07
#769: recompute the normalized value on every run

### LINKED ISSUE #769: `normalize-metrics` fails to recompute `n_*` for late-arriving samples, causing the series to freeze
## What happens

`judges/normalize-metrics/normalize-metrics.rb:40` never revises a value it has written:

```ruby
n = "n_#{prop}"
next if f[n]
f.send(:"#{n}=", diff)
```

The baseline is "the earliest fact that carries this property at the time of the run"
(lines 21-31), and that set grows between runs. `Jp.incremate` fills at most
`qos_max_per_fact` metrics per fact per run and retries metrics that previously returned
`{}`, so a metric really can land on a later-dated fact before an earlier-dated one. When the
earlier value arrives, every `n_*` already written stays frozen against the old baseline:

```
after run 1:
  #0: {"when"=>[2024-01-01]}
  #1: {"when"=>[2024-01-08], "average_issue_lifetime"=>[200.0], "n_average_issue_lifetime"=>[0.0]}

after run 2 (baseline is now 100, so week 2 should read 1.0):
  #0: {"when"=>[2024-01-01], "average_issue_lifetime"=>[100.0], "n_average_issue_lifetime"=>[0.0]}
  #1: {"when"=>[2024-01-08], "average_issue_lifetime"=>[200.0], "n_average_issue_lifetime"=>[0.0]}
```

The metric doubled, and the chart shows a flat line at zero, permanently.

## What should happen

`n_*` should be recomputed on every run, since it is a derived value and its baseline can
move. Overwriting it (via `Fbe.overwrite`) instead of skipping keeps the series consistent
with the data actually present.

morphqdd:
`next if f[n]` meant a normalized value was written once and never revised, but its baseline is
"the earliest fact that carries the property right now", and that set grows between runs. The
collector fills a limited number of metrics per fact per run and retries the ones that came back
empty, so a metric can land on a later-dated fact before an earlier-dated one. When the earlier
value finally arrived, everything already written stayed frozen against the old baseline, and the
chart showed a flat line at zero for a metric that had doubled.

The values are collected per fact and handed to `Fbe.overwrite` in one go, so the fact is
recreated at most once and only when something actually changed.

The new test runs the judge twice, adding the earlier sample in between, and checks the later
week reads `1.0` rather than the `0.0` it was given on the first pass.

Closes #769

### DIFF
--- a/judges/normalize-metrics/normalize-metrics.rb
+++ b/judges/normalize-metrics/normalize-metrics.rb
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: MIT
 
 require 'fbe/fb'
+require 'fbe/overwrite'
 
 def fits?(name)
   return true if name == 'composite'
@@ -30,15 +31,15 @@ end
     end
   end
   facts.each do |f|
+    fresh = {}
     f.all_properties.each do |prop|
       next unless fits?(prop)
       next if start[prop].zero?
       v = f[prop].first.to_f
       s = start[prop]
-      diff = (v - s).to_f / start[prop]
-      n = "n_#{prop}"
-      next if f[n]
-      f.send(:"#{n}=", diff)
+      fresh["n_#{prop}"] = (v - s).to_f / s
     end
+    next if fresh.empty?
+    Fbe.overwrite(f, fresh)
   end
 end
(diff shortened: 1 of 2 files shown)


yegor256:
Merged. Thanks — the two-pass test with the earlier sample arriving in between is exactly the shape that proves it.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6162](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #785, opened 2026-09-07
#773: read the adless flag the same way everywhere

### LINKED ISSUE #773: 'adless' not set correctly results in missing logo but Zerocracy palette and links persist
## What happens

`entry.sh` tests `INPUT_ADLESS` with two different polarities:

```
:181   if [ -z "${logo}" ] && [ "${INPUT_ADLESS}" == 'false' ]; then   # default logo
:194   if [ "${INPUT_ADLESS}" == 'true' ]; then                        # log line
:214   if [ "${INPUT_ADLESS}" != 'true' ]; then                        # palette.css
```

With `INPUT_ADLESS` unset, line 181 is false so the default logo is dropped, while lines 194
and 214 take the ads branch: the run logs "The output HTML will have links to Zerocracy" and
injects `https://www.zerocracy.com/css/palette.css`. The result is an ads-mode page with no
logo.

`INPUT_ADLESS` is unset in `makes/entry-in-docker.sh:35-45` and in 8 of the 17 `entries/*.sh`
scripts. Under GitHub Actions `action.yml:32` supplies the `false` default, so the mismatch
only bites direct and Docker invocations, which is exactly what the `make` targets and the
entry tests use.

## What should happen

One polarity everywhere. `[ "${INPUT_ADLESS}" != 'true' ]` at line 181 makes an unset value
behave the same as `false`, matching lines 194 and 214.

morphqdd:
Line 181 compared the flag to `false` while lines 194 and 214 compare it to `true`, so an unset
`INPUT_ADLESS` produced a page with no logo but with the Zerocracy palette and links, and a log
line saying the output would have those links. The action's own default fills the value in under
GitHub Actions, so this only showed up in direct and Docker runs, which is what the make targets
and most of the entry scripts use.

The new entry script runs the script with the flag unset and checks both halves agree.

Closes #773

### DIFF
--- a/entries/uses-default-logo-without-adless.sh
+++ b/entries/uses-default-logo-without-adless.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+
+set -e -o pipefail
+
+SELF=$1
+
+BUNDLE_GEMFILE="${SELF}/Gemfile"
+export BUNDLE_GEMFILE
+bundle exec judges eval test.fb "\$fb.insert" > /dev/null
+
+env "GITHUB_WORKSPACE=$(pwd)" \
+  'GITHUB_REPOSITORY=foo/bar' \
+  'GITHUB_REPOSITORY_OWNER=foo' \
+  'INPUT_FACTBASE=test.fb' \
+  'INPUT_OUTPUT=output' \
+  'INPUT_VERBOSE=false' \
+  'INPUT_LOGO=' \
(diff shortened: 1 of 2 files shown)


yegor256:
Merged. Thanks — the log line saying one thing while the page did another was the giveaway.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6202](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #786, opened 2026-09-07
#761: let a week keep the award that starts it

### LINKED ISSUE #761: An award at exactly midnight Monday falls into no weekly column
## What happens

`z:in-week` decides which weekly column an award belongs to:

https://github.com/zerocracy/pages-action/blob/master/xsl/awards.xsl

```xml
<xsl:variable name="monday" select="xs:dateTime(z:monday($week))"/>
<xsl:variable name="sunday" select="$monday + xs:dayTimeDuration('P7D')"/>
<xsl:value-of select="xs:dateTime($when) &gt; $monday and xs:dateTime($when) &lt; $sunday"/>
```

`$monday` is midnight of that Monday, and the lower bound is strict, so an award
timestamped exactly at midnight Monday belongs to no column at all. Its own week starts
one instant later, and the week before it ends at the same instant, which the upper bound
also excludes:

```
weeks=4 days=28, today=2026-09-20T12:00:00Z
week 1 monday=2026-08-24 contains 2026-09-07T00:00:00Z: false
week 2 monday=2026-08-31 contains 2026-09-07T00:00:00Z: false
week 3 monday=2026-09-07 contains 2026-09-07T00:00:00Z: false
week 4 monday=2026-09-14 contains 2026-09-07T00:00:00Z: false
```

The row total is computed separately, as `sum($facts/award)` over the whole window, and it
does include the award. So the weekly cells of that row no longer add up to its total, and
`dot.xsl` splits its rows with the same function.

## What should happen

The lower bound should be inclusive (`>=`), so that every award inside the window lands in
exactly one column.

morphqdd:
The lower bound of a week was strict, so an award timestamped at exactly midnight on the Monday
belonged to no column: its own week starts one instant later, and the week before ends at the
same instant, which the upper bound excludes too. The row total is summed over the whole window
and did include the award, so the weekly cells of that row stopped adding up to it. `dot.xsl`
splits its rows with the same function.

The bound is inclusive now, so every fact inside the window lands in exactly one column.

Closes #761

### DIFF
--- a/test/pages/test_awards.rb
+++ b/test/pages/test_awards.rb
@@ -189,6 +189,23 @@ class TestAwards < Minitest::Test
     assert_includes(th['title'], 'Week #1 in 2026', xml)
   end
 
+  def test_fn_in_week_takes_midnight_of_monday
+    xml = xslt(
+      '<r><xsl:value-of select="z:in-week(xs:dateTime(\'2024-09-23T00:00:00Z\'), 2)"/></r>',
+      '
+      <fb>
+        <f>
+          <what>pmp</what>
+          <area>hr</area>
+          <days_of_running_balance>14</days_of_running_balance>
+        </f>
+      </fb>
+      ',
+      'today' => '2024-09-26T04:04:04Z'
+    )
+    assert_equal('true', xml.xpath('/r/text()').to_s, xml)
+  end
+
   def test_fn_award
     {
       42 => ['darkgreen', '+42'],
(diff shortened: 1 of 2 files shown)


yegor256:
Not merged: #789 landed first and this now conflicts with it — both touch the week-boundary logic. The inclusive lower bound is right and I would like to take it; please rebase on `master` and I will.

---

morphqdd:

PR #787, opened 2026-09-07
#758: colour an index against one, not against zero

### LINKED ISSUE #758: CPI and SPI are always green and printed with a `+`, because `z:index` compares them to zero
## What happens

`z:index` paints a number green when it is at or above zero and red below it, and prints
it with a sign:

```xml
<xsl:when test="$i &gt;= 0">
  <xsl:text>darkgreen</xsl:text>
</xsl:when>
...
<xsl:value-of select="z:format-signed($i, '0.00')"/>
```

That is right for the normalized metrics, where zero means "same as the first
measurement". CPI and SPI are not those: they are ratios whose neutral point is one, and
`eva-summary.xsl` prints them through the same function:

```xml
<xsl:copy-of select="z:index(ev div ac)"/>
<xsl:copy-of select="z:index(ev div pv)"/>
```

So a project that has spent five times what it earned is shown in green:

```
AC: 1000, EV: 200, PV: 900
-> AC: 1000, EV: 200, PV: 900, CPI: <span class="darkgreen">+0.20</span>,
   SPI: <span class="darkgreen">+0.22</span>.
```

CPI 0.20 means the money is going five times slower than the value, and the page says it
in green with a plus in front of it. Both indexes are always green, since a ratio of two
non-negative numbers is never below zero.

## What should happen

CPI and SPI should be colored against one, not against zero, and printed without the
sign, since a ratio is not a delta.

morphqdd:
`z:index` painted a number green at or above zero and printed it with a sign. Its only two
callers are CPI and SPI, which are ratios of two non-negative amounts: they are never below
zero, so both were always green, and a project that had spent five times what it earned was
shown as `CPI: +0.20` in green.

One is the neutral point of a ratio, so that is what the colour compares against, and the sign
is gone because a ratio is not a delta.

Closes #758

### DIFF
--- a/test/pages/test_vitals.rb
+++ b/test/pages/test_vitals.rb
@@ -28,9 +28,10 @@ class TestVitals < Minitest::Test
 
   def test_fn_index
     {
-      3.3 => ['darkgreen', '+3.30'],
-      0 => ['darkgreen', '+0.00'],
-      -1 => ['darkred', '-1.00']
+      3.3 => ['darkgreen', '3.30'],
+      1 => ['darkgreen', '1.00'],
+      0.2 => ['darkred', '0.20'],
+      0 => ['darkred', '0.00']
     }.each do |k, v|
       xml = xslt("<xsl:copy-of select='z:index(#{k})'/>", '<fb/>')
       assert_equal(v[0], xml.xpath('/span/@class').to_s, xml)
--- a/xsl/vitals.xsl
+++ b/xsl/vitals.xsl
@@ -47,14 +47,16 @@
   </xsl:function>
   <xsl:function name="z:index">
     <!--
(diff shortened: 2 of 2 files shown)


yegor256:
Merged. Thanks — a ratio is not a delta, and both callers really were always green.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (17) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6242](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #789, opened 2026-09-07
#760: read the moment of a fact that carries more than one

### LINKED ISSUE #760: One fact with two `when` values fails the whole page with `FORG0001`
## What happens

A property of a fact is a set of values, so a fact may carry two `when` values, and
`Factbase::ToXML` prints such a property as a container:

```xml
<when>
  <v t="T">2026-09-01T00:00:00Z</v>
  <v t="T">2026-09-02T00:00:00Z</v>
</when>
```

Every stylesheet here reads `when` as a single value, for example:

https://github.com/zerocracy/pages-action/blob/master/xsl/qo-section.xsl

```xml
<xsl:variable name="raw" select="/fb/f[what=$what and xs:dateTime(when) &gt; (...)]"/>
```

so the whole page fails to build:

```
Error at char 37 in xsl:variable/@select on line 28 column 137 of qo-section.xsl:
  FORG0001: Invalid dateTime value "2026-09-01T00:00:00Z2026-09-02..." (Characters after the end)
```

Saxon exits non-zero, `entry.sh` runs under `set -e`, and the action produces no HTML at
all. The message names neither the fact nor the property, so the reason has to be guessed
from the concatenated string.

One fact like that is enough, and it can come from any judge that writes `when` twice,
including a judge in the user's own repository, since this action renders any factbase.

## What should happen

A fact whose `when` (or any other property the stylesheets read as a scalar) has more than
one value should not take the whole page down. Reading the first value, or skipping the
fact with a message naming it, would both be better than an XPath type error.

morphqdd:
A property of a fact is a set of values, so `when` may hold two, and the XML then carries them as
`v` children. Every stylesheet read it as a scalar, so `xs:dateTime(when)` got the two timestamps
glued together and raised `FORG0001`. Saxon exits non-zero, the entry script runs under `set -e`,
and the action produced no HTML at all, with a message naming neither the fact nor the property.
One such fact was enough, and it can come from any judge, including one in the user's own
repository.

`z:when` reads that property in one place and answers with the earliest value it holds, so a fact
written twice lands in one week instead of taking the page down. Every reader goes through it,
and `z:in-week` now takes the moment rather than the node, so the two cannot drift apart.

`badge.xsl` and `assessment.xsl` gained the `z` namespace, which they did not declare before.

Closes #760

### DIFF
--- a/test/pages/test_awards.rb
+++ b/test/pages/test_awards.rb
@@ -148,7 +148,7 @@ class TestAwards < Minitest::Test
 
   def test_fn_in_week
     xml = xslt(
-      '<r><xsl:value-of select="z:in-week(\'2024-09-20T04:04:04Z\', 1)"/></r>',
+      '<r><xsl:value-of select="z:in-week(xs:dateTime(\'2024-09-20T04:04:04Z\'), 1)"/></r>',
       '
       <fb>
         <f>
--- a/test/pages/test_vitals.rb
+++ b/test/pages/test_vitals.rb
@@ -38,6 +38,37 @@ class TestVitals < Minitest::Test
     end
   end
 
+  def test_renders_a_fact_that_has_two_moments
+    xml = xslt(
+      "<r><xsl:apply-templates select='/' mode='awards'/></r>",
+      '
+      <fb>
+        <f>
+          <what>pmp</what>
+          <area>hr</area>
+          <days_of_running_balance>7</days_of_running_balance>
+        </f>
+        <f>
+          <what>something</what>
+          <who>5</who>
+        </f>
+        <f>
+          <is_human>1</is_human>
+          <who>1</who>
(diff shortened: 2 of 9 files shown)


yegor256:
Merged. Thanks — one fact written twice taking the whole page down was worth fixing at the root, and routing every reader through `z:when` is the right place for it.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points for this: +24 as a basis; +7 for the 100 hits-of-code that you wrote; -8 to not go over the cap. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6138](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #803, opened 2026-09-07
#797: keep the bylaws until there are new ones

### LINKED ISSUE #797: `add-bylaws-html` prematurely deletes bylaws without confirming ability to rebuild
## What happens

`add-bylaws-html` throws away what it has before it knows whether it can produce a replacement:

https://github.com/zerocracy/pages-action/blob/master/judges/add-bylaws-html/add-bylaws-html.rb

```ruby
Fbe.fb.query('(eq what "bylaws")').delete!
return if htmls.empty?
```

`htmls` is empty whenever the `pmp/hr` fact is missing, and also whenever it is there but none of
its properties start with `(award`, which is the ordinary case for a project that configures a
running balance and no award formulas yet. Either way the bylaws that a previous run built are
gone, and `bylaws.xsl` falls back to the red "There is no information about bylaws."

```
$ ruby t_bylaws.rb
after run 1: bylaws=1
after a pmp/hr with no award formula: bylaws=0
```

Same result when the `pmp/hr` fact is removed entirely.

The sibling judge gets the order right, which makes the contrast easy to see:

https://github.com/zerocracy/pages-action/blob/master/judges/latest-assessment/latest-assessment.rb

```ruby
assessments = Fbe.fb.query('(and (eq what "assessment") ...)').each.to_a
return if assessments.empty?
...
Fbe.fb.query('(eq what "latest-assessment")').delete!
```

## What should happen

The judge should keep the bylaws it has until it has new ones to put in their place, the way
`latest-assessment` does: leave early when there is nothing to write, and only then delete and
insert.

morphqdd:
The judge deleted the old bylaws before it knew whether it had anything to put in their place, so a missing `pmp/hr` fact, or one whose properties carry no `(award` formula, wiped what the previous run had built and left the page showing "There is no information about bylaws."

The delete now happens after the empty check, which is the order `latest-assessment` already uses.

Closes #797

### DIFF
--- a/judges/add-bylaws-html/add-bylaws-html.rb
+++ b/judges/add-bylaws-html/add-bylaws-html.rb
@@ -18,8 +18,8 @@ f&.all_properties&.each do |prop|
   htmls << Redcarpet::Markdown.new(Redcarpet::Render::HTML.new(escape_html: true)).render(md)
   par += 1
 end
-Fbe.fb.query('(eq what "bylaws")').delete!
 return if htmls.empty?
+Fbe.fb.query('(eq what "bylaws")').delete!
 s = Fbe.fb.insert
 s.what = 'bylaws'
 s.html = htmls.join
--- a/test/judges/test_add_bylaws_html.rb
+++ b/test/judges/test_add_bylaws_html.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+
+require 'factbase'
+require_relative '../test__helper'
+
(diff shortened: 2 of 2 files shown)


morphqdd:
@yegor256, CI is green here. Please take a look.

yegor256:
Merged, thanks! Matching the order `latest-assessment` already uses is the right precedent to follow.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+7946](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #815, opened 2026-09-08
#801: refuse a timeout or lifetime that is not a positive integer

### LINKED ISSUE #801: `entry.sh` wrongly transforms non-numeric `INPUT_TIMEOUT` and `INPUT_LIFETIME` values into zero
The `entry.sh` conversions at lines 146-157 multiply `INPUT_TIMEOUT` and `INPUT_LIFETIME` directly in Bash arithmetic without validating that either value is numeric. Bash treats an unrecognized value such as `abc` as zero, so `INPUT_TIMEOUT=abc` becomes `--timeout 0` and `INPUT_LIFETIME=abc` becomes `--lifetime 0` while the action continues.

The subsequent `judges update` therefore runs with zero-second budgets and can generate an incomplete page instead of rejecting the configuration. Validate both inputs as positive integer minute counts before arithmetic expansion and fail with a clear message.

gemshrine:
Fixes #801

`INPUT_TIMEOUT` and `INPUT_LIFETIME` went straight into `$(( ))`, and Bash reads an unrecognized word there as zero. `INPUT_TIMEOUT=abc` became `--timeout 0`, the run went ahead with a zero-second budget, and the action published whatever page it had instead of saying the input was wrong.

Both are checked against `^[1-9][0-9]*$` before the arithmetic now, so `abc`, `0`, `-5` and `1.5` stop the run with a message naming the input. The empty case is unchanged: 3 and 5 minutes as before.

The new entry test drives the real `entry.sh` four times, once per bad value per input, and requires each run to exit non-zero with that message. On master every one of those runs exits 0.


yegor256:
Merged, thanks! Driving the real `entry.sh` four times makes the guard genuinely covered.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2208](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #833, opened 2026-09-08
Avoid duplicate script template includes

### LINKED ISSUE #807: Multiple inclusions of script-with-cdata.xsl causes Saxon warnings during rendering
## What happens

`script-with-cdata.xsl` is pulled into the vitals build three times:

```
xsl/vitals.xsl:7      <xsl:include href="script-with-cdata.xsl"/>
xsl/awards.xsl:7      <xsl:include href="script-with-cdata.xsl"/>
xsl/qo-section.xsl:7  <xsl:include href="script-with-cdata.xsl"/>
```

and `vitals.xsl` imports both of the other two:

```
xsl/vitals.xsl:23     <xsl:import href="awards.xsl"/>
xsl/vitals.xsl:27     <xsl:import href="qo-section.xsl"/>
```

Saxon says so on every render:

```
Warning at xsl:stylesheet on line 6 column 80 of script-with-cdata.xsl:
  SXWN9019  Stylesheet module .../xsl/script-with-cdata.xsl is included or
  imported more than once. This is permitted, but may lead to errors or unexpected behavior
```

## Reproduction

Any run of the real stylesheet, no special input needed:

```
java -cp saxon.jar:xmlresolver.jar net.sf.saxon.Transform \
  -s:factbase.xml -xsl:xsl/vitals.xsl today=2026-09-07T12:00:00Z
```

The warning comes out before the first line of HTML. `entry.sh:252` runs exactly this, so the message is in every build log the action produces.

## Why it matters

The named template `script-with-cdata` ends up defined at two different import precedences: once directly in `vitals.xsl`, which is the top of the tree, and once inside `awards.xsl` and `qo-section.xsl`, which `vitals.xsl` imports and which therefore sit lower. XSLT resolves a name by precedence, so which definition a call binds to depends on where the call is written, not on which module it looks like it is calling. Today all three copies are the same file, so nothing breaks; the day somebody edits the template with one call site in mind, the other call sites may or may not follow, and the failure will look like the template ignoring the change.

The warning is also noise that trains readers to skip warnings in the build log, which is where the next real one will appear.

## Suggested fix

Include it once, at the top:

```
xsl/vitals.xsl        keeps  <xsl:include href="script-with-cdata.xsl"/>
xsl/awards.xsl        drops it
xsl/qo-section.xsl    drops it
```

`awards.xsl` and `qo-section.xsl` are only ever reached through `vitals.xsl`, both in `entry.sh` and in `test/test__helper.rb`, which imports `vitals.xsl` and applies templates into it, so neither module needs its own copy.

`when.xsl` is included twice as well, by `vitals.xsl` and by `badge.xsl`, but those are two separate top-level stylesheets run by two separate Saxon invocations, so that one is correct as it is.

After the change the render is silent, which makes the next warning worth reading.

gemshrine:
## What's changed

`script-with-cdata.xsl` is now included only by the top-level `vitals.xsl` stylesheet. The redundant includes were removed from `awards.xsl` and `qo-section.xsl`.

## Why

Both modules are imported by `vitals.xsl`, so the template was loaded three times for every vitals render. Saxon emitted an `SXWN9019` warning and maintained multiple import-precedence copies of the same named template. Keeping one top-level include removes the warning and leaves one unambiguous definition for all call sites.

## Verification

- `git diff --check`
- Confirmed both modules are imported through `vitals.xsl`, which retains the include.

Closes #807

### DIFF
--- a/js/outdated.js
+++ b/js/outdated.js
@@ -58,22 +58,18 @@ function formatRelativeTime(diffInMs, startDate) {
  * Displays a warning message if the page is outdated
  */
 function displayOutdatedWarning() {
-  const time = Date.parse($("#generated-time").attr("datetime"));
-  if (isNaN(time)) {
-    console.error("Could not parse the generated time");
-    return;
-  }
-  const hours = Math.floor((Date.now() - time) / TIME_UNITS.HOUR);
-  const warning = $("#page-outdated-warning");
-  if (hours > OUTDATED_THRESHOLD_HOURS) {
-    const message = `This page was generated ${hours} ${hours === 1 ? 'hour' : 'hours'} ago. The information is most probably outdated.`;
-    if (warning.length === 0) {
-      $('footer').prepend(`<p id="page-outdated-warning" class='red'>${message}</p>`);
-    } else {
-      warning.text(message);
+  if ($("#page-outdated-warning").length === 0) {
+    const time = Date.parse($("#generated-time").attr("datetime"));
+    if (isNaN(time)) {
+      console.error("Could not parse the generated time");
+      return;
     }
-  } else if (warning.length > 0) {
-    warning.remove();
+    const hours = Math.floor((Date.now() - time) / TIME_UNITS.HOUR);
+    if (hours > OUTDATED_THRESHOLD_HOURS) {
+      $('footer').prepend(
+        `<p id="page-outdated-warning" class='red'>This page was generated ${hours} ${hours === 1 ? 'hour' : 'hours'} ago. The information is most probably outdated.</p>`
+      );
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


yegor256:
Merged, thanks! One unambiguous definition beats three with import precedence between them.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2184](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #834, opened 2026-09-08
Refresh outdated page warning text

### LINKED ISSUE #806: Outdated-page banner doesn't update timestamp after initial display
## What happens

`displayOutdatedWarning` refuses to do anything once the banner is on the page:

```js
function displayOutdatedWarning() {
  if ($("#page-outdated-warning").length === 0) {
    const time = Date.parse($("#generated-time").attr("datetime"));
    ...
    const hours = Math.floor((Date.now() - time) / TIME_UNITS.HOUR);
    if (hours > OUTDATED_THRESHOLD_HOURS) {
      $('footer').prepend(
        `<p id="page-outdated-warning" ...>This page was generated ${hours} hours ago. ...</p>`
      );
    }
  }
}
```

The guard is there to avoid a second banner, but it also skips the recalculation, so the number is written once and never touched again. Meanwhile the timer next to it keeps running:

```js
setInterval(function() {
  updateTimeDisplay();
  displayOutdatedWarning();
}, TIME_UNITS.MINUTE);
```

`updateTimeDisplay` rewrites every `time.relative-time` element every minute, correctly. `displayOutdatedWarning` returns immediately.

## Reproduction

Driving the function with a stub `$`, a page generated at `2026-09-01T00:00:00Z`, and the clock moved forward:

```
real age   25h -> banner says 25h
real age   26h -> banner says 25h
real age   48h -> banner says 25h
real age  168h -> banner says 25h
real age  720h -> banner says 25h
```

A page left open for a month keeps insisting it was generated 25 hours ago.

## Why it matters

The banner exists for exactly the case where generation has stopped: the workflow is broken, the token expired, the schedule was disabled. That is when somebody keeps the vitals page open and watches how stale it gets. The relative timestamps in the table climb correctly while the banner stays at whatever it read the first time, so the two disagree with each other on the same screen, and the banner is the one that is wrong.

The 24-hour threshold makes it worse: the first reading is always just above 24, so the frozen number is always a small one. The page never says "generated 6 days ago", it says "25 hours ago" forever.

There is a smaller thing in the same line. The plural is chosen with `hours === 1 ? 'hour' : 'hours'`, inside a branch that only runs when `hours > 24`, so the singular can never be printed.

## Suggested fix

Update the banner instead of skipping it:

```js
function displayOutdatedWarning() {
  const time = Date.parse($("#generated-time").attr("datetime"));
  if (isNaN(time)) {
    console.error("Could not parse the generated time");
    return;
  }
  const hours = Math.floor((Date.now() - time) / TIME_UNITS.HOUR);
  if (hours <= OUTDATED_THRESHOLD_HOURS) {
    $("#page-outdated-warning").remove();
    return;
  }
  const text = `This page was generated ${hours} hours ago. The information is most probably outdated.`;
  const existing = $("#page-outdated-warning");
  if (existing.length === 0) {
    $('footer').prepend(`<p id="page-outdated-warning" class='red'>${text}</p>`);
  } else {
    existing.text(text);
  }
}
```

The `remove()` branch covers the other direction, which the current code cannot do either: a page that is refreshed by the browser while the tab is open keeps the banner even after a fresh generation.

Since the message is now built once, the dead singular branch goes away with it.

gemshrine:
## What's changed

The outdated-page warning is recalculated on every timer tick. Its text is updated as the page ages, and the warning is removed if the page becomes fresh again.

## Why

The previous function returned as soon as the warning existed. A page that stayed open therefore displayed the age from the first check forever, while the relative timestamps around it continued changing. The warning could also remain after a refreshed page became current.

## Verification

- `git diff --check`
- The existing one-minute timer now updates or removes the existing warning element instead of returning early.

Closes #806

### DIFF
--- a/js/outdated.js
+++ b/js/outdated.js
@@ -58,18 +58,22 @@ function formatRelativeTime(diffInMs, startDate) {
  * Displays a warning message if the page is outdated
  */
 function displayOutdatedWarning() {
-  if ($("#page-outdated-warning").length === 0) {
-    const time = Date.parse($("#generated-time").attr("datetime"));
-    if (isNaN(time)) {
-      console.error("Could not parse the generated time");
-      return;
+  const time = Date.parse($("#generated-time").attr("datetime"));
+  if (isNaN(time)) {
+    console.error("Could not parse the generated time");
+    return;
+  }
+  const hours = Math.floor((Date.now() - time) / TIME_UNITS.HOUR);
+  const warning = $("#page-outdated-warning");
+  if (hours > OUTDATED_THRESHOLD_HOURS) {
+    const message = `This page was generated ${hours} ${hours === 1 ? 'hour' : 'hours'} ago. The information is most probably outdated.`;
+    if (warning.length === 0) {
+      $('footer').prepend(`<p id="page-outdated-warning" class='red'>${message}</p>`);
+    } else {
+      warning.text(message);
     }
-    const hours = Math.floor((Date.now() - time) / TIME_UNITS.HOUR);
-    if (hours > OUTDATED_THRESHOLD_HOURS) {
-      $('footer').prepend(
-        `<p id="page-outdated-warning" class='red'>This page was generated ${hours} ${hours === 1 ? 'hour' : 'hours'} ago. The information is most probably outdated.</p>`
-      );
-    }
+  } else if (warning.length > 0) {
+    warning.remove();
   }
 }
 


yegor256:
Merged, thanks! Removing the warning when the page becomes fresh again is a good addition.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (26) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2164](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #835, opened 2026-09-08
Validate today input before judges run

### LINKED ISSUE #811: `today` input without full ISO-8601 date-time causes failure in final rendering
## What happens

`entry.sh` takes the `today` input and hands it to Saxon without looking at it:

```bash
if [ -z "${INPUT_TODAY}" ]; then
    INPUT_TODAY=$(date -u +'%Y-%m-%dT%H:%M:%SZ')
else
    echo "The 'today' is set to: '${INPUT_TODAY}'"
fi
...
java -jar "${SELF}/target/saxon.jar" ... "today=${INPUT_TODAY}" ...
```

`vitals.xsl` declares it as `xs:string` and casts it later with `xs:dateTime($today)`, so a value that is not a full ISO-8601 date-time blows up inside the stylesheet.

## Reproduction

Same factbase, same stylesheet, only `today` changed:

| `today` | result |
| --- | --- |
| `2026-09-07T12:00:00Z` | renders |
| `2026-09-07T12:00:00` | renders |
| `2026-09-07` | `FORG0001 Invalid dateTime value "2026-09-07" (Too short)` |
| `2026-09-07 12:00:00Z` | `FORG0001 Invalid dateTime value "2026-09-07 12:00:00Z" (Day must be two digits)` |

## Why it matters

The failure lands at the very end. By the time Saxon runs, `entry.sh` has already run `judges update` across all six judges, spent the GitHub quota, written the factbase, and produced `test.xml`, `test.yaml` and `test.json`. The HTML and the badge are the only things left, and the run dies there, so the workflow fails after doing all of the work and none of the publishing.

The message does not name the input. Somebody reading `FORG0001 Invalid dateTime value` in a build log has to know that it comes from the `today:` parameter of the action, because nothing in the error says so.

For the space-separated form the message is worse than unhelpful, it is wrong: `Day must be two digits` about a value whose day is `07`.

Writing `today: 2026-09-07` is the natural mistake, because `action.yml` says "ISO-8601 date-time" and a bare date is valid ISO-8601. The action's own default is a full date-time, so the only people who hit this are the ones who set the input by hand.

## Suggested fix

Check the shape where the value arrives, next to the check `cycles` already gets in the sibling action:

```bash
if [ -n "${INPUT_TODAY}" ]; then
    if ! [[ "${INPUT_TODAY}" =~ ^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}(\.[0-9]+)?(Z|[+-][0-9]{2}:[0-9]{2})?$ ]]; then
        echo "The 'today' input must be an ISO-8601 date-time, like 2026-09-07T12:00:00Z, but it is: '${INPUT_TODAY}'" >&2
        exit 1
    fi
    echo "The 'today' is set to: '${INPUT_TODAY}'"
fi
```

Failing before `judges update` costs the user a few seconds instead of a whole run, and the message names the parameter.

Accepting a bare date and appending `T00:00:00Z` to it would be friendlier still, but the check is the part that matters.

gemshrine:
## What's changed

The optional `today` input is validated as an ISO-8601 date-time before any judges are executed. Invalid values now produce an actionable error naming the input.

## Why

A bare date or another incomplete value previously passed through `entry.sh`, and Saxon failed only after the factbase update and output generation had already completed. This wastes the run and leaves the build log with an opaque `FORG0001` error.

## Verification

- `bash -n entry.sh`
- `git diff --check`
- Validation runs before `judges update`.

Closes #811

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -143,6 +143,16 @@ if [ "${github_token_found}" == "false" ]; then
     fi
 fi
 
+if [ -z "${INPUT_TODAY}" ]; then
+    INPUT_TODAY=$(date -u +'%Y-%m-%dT%H:%M:%SZ')
+else
+    if ! [[ "${INPUT_TODAY}" =~ ^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}(\.[0-9]+)?(Z|[+-][0-9]{2}:[0-9]{2})?$ ]]; then
+        echo "ERROR: The 'today' input must be an ISO-8601 date-time, like 2026-09-07T12:00:00Z, but it is: '${INPUT_TODAY}'" >&2
+        exit 1
+    fi
+    echo "The 'today' is set to: '${INPUT_TODAY}'"
+fi
+
 timeout=${INPUT_TIMEOUT}
 if [ -z "${timeout}" ]; then
     timeout=3
@@ -170,12 +180,6 @@ ${JUDGES} "${gopts[@]}" print \
     --format xml \
     "${INPUT_FACTBASE}" \
     "${INPUT_OUTPUT}/${name}.rich.xml"
-
-if [ -z "${INPUT_TODAY}" ]; then
-    INPUT_TODAY=$(date -u +'%Y-%m-%dT%H:%M:%SZ')
-else
-    echo "The 'today' is set to: '${INPUT_TODAY}'"
-fi
 
 logo=${INPUT_LOGO}
 if [ -z "${logo}" ] && [ "${INPUT_ADLESS}" != 'true' ]; then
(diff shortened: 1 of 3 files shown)


yegor256:
Merged, thanks! Failing before the update rather than after it saves the whole run.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2144](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #836, opened 2026-09-08
Select payables reconciliation by timestamp

### LINKED ISSUE #800: `z:payables` function selects reconciliation record by XML order, not latest `when` timestamp
The `z:payables` function in `xsl/awards.xsl:43` selects the reconciliation record with `[last()]` after filtering by programmer. That is the last node in the input XML, not the record with the greatest `when` timestamp.

If reconciliation facts are imported, merged, or otherwise stored out of chronological order, the Pay column uses an older payout (`awarded`, `balance`, and `since`) and shows a wrong payable amount even though a newer reconciliation is present. The page already treats `when` as the timestamp that determines recency elsewhere.

Select the reconciliation with the maximum `when` value (for example, sort matching records by `xs:dateTime(when)` descending and take the first) before calculating the accumulated and payable values.

gemshrine:
## What's changed

`z:payables` now selects the reconciliation fact with the greatest `when` timestamp, regardless of the order in which facts appear in the XML. A regression test covers out-of-order reconciliation facts.

## Why

The previous `[last()]` expression selected the last matching XML node, not the latest reconciliation. Imported or merged facts can arrive out of chronological order, causing the Pay column to use stale `awarded`, `balance`, and `since` values.

## Verification

- `bundle exec ruby -Itest test/pages/test_awards.rb` (9 tests, 17 assertions)
- `git diff --check`

Closes #800


opusstudiohq-max:
I had opened #840 for the same defect before spotting this one — yours came first, so I've closed mine. One observation from writing it that may be worth folding in here.

The cast reads the property directly:

```xml
<xsl:variable name="latest" select="max(for $reconciliation in $recs return xs:dateTime($reconciliation/when))"/>
```

A fact's property is a set of values, so a `when` written twice arrives as several `<v>` children rather than as text, and `xs:dateTime` on that raises a type error. `xsl/when.xsl` exists for exactly this case, and says so:

> The earliest of them is the answer, so that a fact written twice still lands in one place instead of taking the whole page down with a type error.

The accumulation a few lines below already goes through it — `z:when(when) > xs:dateTime($rec/since)` — so `z:when(when)` in place of `xs:dateTime(...)/when` here would keep the two readings consistent and leave a duplicated `when` a wrong-number problem rather than a broken-page one.

Not a blocker for the defect you're fixing, which is real and correctly diagnosed.

gemshrine:
Thanks for the careful review and for closing the duplicate PR. I updated #836 to use the shared z:when() helper when selecting and rendering the latest reconciliation, and added coverage for repeated when values. The change is pushed to this PR.

yegor256:
Merged, thanks! `last()` reading document order has bitten several of these stylesheets.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +24 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2124](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #838, opened 2026-09-08
Mask GitHub tokens before command tracing

### LINKED ISSUE #813: Verbose mode exposes unmasked GitHub token in the command trace
## What happens

With `verbose: true`, `entry.sh` turns bash tracing on at line 32 and leaves it on for the rest of the run:

```bash
if [ "${INPUT_VERBOSE}" == 'true' ]; then
    set -x
fi
```

The token is appended to the options with tracing deliberately switched off around it:

```bash
set +x
options+=("--option=github_token=$(printenv "INPUT_GITHUB-TOKEN")");
[ "${INPUT_VERBOSE}" == 'true' ] && set -x
```

That guard covers the assignment. It does not cover the line that uses the array, sixty lines below, which runs with tracing back on:

```bash
${JUDGES} "${gopts[@]}" update \
    ...
    "${options[@]}" \
    "${SELF}/judges/" "${INPUT_FACTBASE}"
```

`set -x` prints the fully expanded command, so the token is printed there, in full.

## Reproduction

The same structure, reduced:

```bash
set -e -o pipefail
INPUT_VERBOSE=true
if [ "${INPUT_VERBOSE}" == 'true' ]; then set -x; fi
declare -a options=()
set +x
options+=("--option=github_token=${TOKEN}")
[ "${INPUT_VERBOSE}" == 'true' ] && set -x
echo "judges update" "${options[@]}" > /dev/null
```

stderr:

```
+ echo 'judges update' --option=github_token=ghp_SUPERSECRET123
```

The assignment is invisible, the use is not.

## Why it matters

`entry.sh` here registers no workflow masks at all. Its sibling in `zerocracy/judges-action` does, with a comment that says exactly why:

```bash
# Mask token values in CI logs before enabling bash tracing.
# Without these workflow commands, "set -x" emits the full command line
# including "--option=github_token=ghp_..." and "--token=ZRCY-..." into
# Actions logs, which are readable by anyone with read access to the repo.
if [ -n "$(printenv "INPUT_GITHUB-TOKEN")" ]; then
    echo "::add-mask::$(printenv "INPUT_GITHUB-TOKEN")"
fi
```

That reasoning applies here word for word, and the guard was never copied over. The `set +x` around the assignment reads like the same protection, which is probably why nobody looked further.

How much is exposed depends on where the token comes from. GitHub masks anything that flows out of `secrets.*` on its own, and `github-token` defaults to `${{ github.token }}`, so the common configuration is covered by the platform rather than by this script. A token that is not a secret — one from `vars`, one minted by an earlier step, one passed through `options:` as `github_token=...`, which `entry.sh` explicitly supports — is printed in the clear in a log that anyone with read access to the repository can open.

## Suggested fix

Register the mask before tracing is enabled, the way the sibling action does:

```bash
if [ -n "$(printenv "INPUT_GITHUB-TOKEN")" ]; then
    echo "::add-mask::$(printenv "INPUT_GITHUB-TOKEN")"
fi
for opt in "${options[@]}"; do
    if [[ "${opt}" == "--option=github_token="* ]]; then
        echo "::add-mask::${opt#--option=github_token=}"
    fi
done
```

placed above the `set -x` at line 32 for the first, and right after the options are parsed for the second. Once the value is masked, the `set +x` dance around the assignment can go, because a masked value stays masked wherever it appears.

Note that `[ "${INPUT_VERBOSE}" == 'true' ] && set -x` as the last statement of that branch is also the shape shellcheck flags as SC2015; writing it as an `if` removes the question of what its exit status does under `set -e`.

gemshrine:
## What's changed

GitHub Actions mask commands are emitted for the `github-token` input before verbose tracing is enabled, and for `github_token=` values found in the options list before those options are expanded.

## Why

With `verbose: true`, Bash tracing prints the expanded `judges update` command. The previous `set +x` guard hid only the token assignment; the later command expansion still exposed the token in logs. Tokens supplied through `INPUT_OPTIONS` had the same risk.

Masking happens while tracing is disabled for the mask command itself, then tracing may safely resume because GitHub redacts the registered values wherever they appear in the log.

## Verification

- `bash -n entry.sh`
- `git diff --check`

Closes #813

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -28,6 +28,11 @@ fi
 
 echo "The 'pages-action' ${VERSION} is running"
 
+github_token=$(printenv "INPUT_GITHUB-TOKEN" || true)
+if [ -n "${github_token}" ]; then
+    echo "::add-mask::${github_token}"
+fi
+
 if [ "${INPUT_VERBOSE}" == 'true' ]; then
     set -x
 fi
@@ -118,6 +123,15 @@ while IFS= read -r o; do
     v="${v%"${v##*[![:space:]]}"}"
     if [ "${v}" = "" ]; then
         continue
+    fi
+    if [[ "${v}" == "github_token="* ]]; then
+        if [ "${INPUT_VERBOSE}" == 'true' ]; then
+            set +x
+        fi
+        echo "::add-mask::${v#github_token=}"
+        if [ "${INPUT_VERBOSE}" == 'true' ]; then
+            set -x
+        fi
     fi
     options+=("--option=${v}")
 done <<< "${INPUT_OPTIONS}"


yegor256:
Merged, thanks! Registering the mask while tracing is off, then letting it resume, is the right order.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +20 points for this: +24 as a basis; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2100](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #839, opened 2026-09-08
Pin external asset integrity hashes

### LINKED ISSUE #812: Subresource Integrity: CDN-served files' hashes are computed at build time, not pinned in advance
## What happens

`entry.sh` builds the `integrity="sha384-..."` attributes by downloading each CDN file during the build and hashing whatever came back:

```bash
for css in "${css_urls[@]}"; do
    hash=$(curl -sSf --max-time 30 "$css" | openssl dgst -sha384 -binary | openssl base64 -A) || { ... }
    css_links="${css_links}${css}|${hash}"$'\n'
done
```

and the same loop for the four JS URLs a few lines below.

Subresource Integrity works by pinning a hash that the publisher decided on **in advance**, so that a browser can refuse a file that does not match it. A hash taken from the file the CDN happens to serve at build time cannot refuse anything: whatever arrives is, by construction, what the hash says. If a CDN serves a modified `jquery.min.js` during the build, the action computes the hash of the modified file and publishes it as the verified one, and every visitor's browser then confirms that the tampered file is exactly the file the page expects.

The attribute is descriptive, not prescriptive. It protects the page against a CDN that changes the file **after** the build, and against nothing else.

## How to see it

`https://www.zerocracy.com/css/palette.css` is in the list without a version, so its content can change between two builds. Each build produces a different `integrity` value for the same URL, and both builds pass. A pinned hash would have failed the second build, which is the whole point of pinning.

The other five URLs are version-pinned, so their hashes happen to be stable today. That is a property of those URLs, not of the mechanism.

## Why it matters

Six third-party files are loaded into every published dashboard: `tacit-css`, `drops`, `palette.css`, `jquery`, `jquery.tablesorter`, `chart.js` and `chroma-js`. The page carries `integrity` attributes for all of them, which reads as "this action pins its dependencies", and a reviewer has no reason to look closer. What is actually pinned is "whatever the CDN said at build time".

Also, `curl -sSf` failing makes the whole action exit 1, so a CDN being briefly unreachable stops a dashboard from being published at all. That is the cost side of a mechanism that is not buying protection.

## Suggested fix

Write the hashes down, next to the URLs, and fail when they do not match:

```bash
declare -A css_urls=(
    ["https://cdn.jsdelivr.net/npm/tacit-css@1.9.5/dist/tacit-css.min.css"]="sha384-<pinned>"
    ["https://cdn.jsdelivr.net/npm/drops@0.3.2/dist/drops-0.3.2.min.css"]="sha384-<pinned>"
)
for css in "${!css_urls[@]}"; do
    got=$(curl -sSf --max-time 30 "$css" | openssl dgst -sha384 -binary | openssl base64 -A)
    if [ "sha384-${got}" != "${css_urls[$css]}" ]; then
        echo "ERROR: ${css} does not match its pinned hash" >&2
        exit 1
    fi
done
```

Renovate already updates the pinned versions in this repo, and a hash next to the version is the same kind of change to review.

`palette.css` needs a version in its URL before it can be pinned at all, or it should be vendored into the image, since it is served from Zerocracy's own host and does not need to be a third-party subresource.

If the hashes stay computed, the honest thing is to drop the `integrity` attributes, because an attribute that always matches tells the reader something that is not true.

## Related

#581 lists the missing JS integrity attributes as a supply-chain gap and points at the CSS loop as the pattern to copy. The attributes have since been added for JS. This issue is about the pattern itself: copying it to more URLs widened the appearance of protection without adding any.

gemshrine:
## What's changed

The action now keeps SHA-384 values next to each external CSS and JavaScript URL and passes those pinned values into the generated page. Build-time downloads and hash calculation have been removed.

## Why

Computing a hash from whatever a CDN serves during the build makes the resulting SRI attribute accept that same response, including a modified response. It does not provide a pre-existing integrity decision. A pinned value makes a changed asset fail the browser's integrity check and turns dependency changes into reviewable source changes.

The existing versioned URLs retain their versions, and the Zerocracy palette URL is pinned to its current published content as well.

## Verification

- `bash -n entry.sh`
- `git diff --check`
- Hashes were calculated from the current contents of all seven configured assets.

Closes #812

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -211,17 +211,19 @@ declare -a css_urls=(
     "https://cdn.jsdelivr.net/npm/tacit-css@1.9.5/dist/tacit-css.min.css"
     "https://cdn.jsdelivr.net/npm/drops@0.3.2/dist/drops-0.3.2.min.css"
 )
+declare -a css_hashes=(
+    "QF+7u3GMRHdbU5nzxyNvTNkV0xQtQXEbAraWZcussxwNUsY3zrmDeAKc8jZ5jfTb"
+    "p1uPBv+9Vp2mQmUJHUrsDnDVbg88Upi66Gf8uTcVy0I2ZIHbSUIEj3q+S6D7B1Lz"
+)
 if [ "${INPUT_ADLESS}" != 'true' ]; then
     css_urls+=("https://www.zerocracy.com/css/palette.css")
+    css_hashes+=("qHLQECQpsl5ded8ztb55ayHqxsGIT9cLHxPH1jOszgf/0HsckECmdIgvAp/PgwZ0")
 fi
 css_links=""
-for css in "${css_urls[@]}"; do
-    echo "Calculating hash for: ${css}"
-    hash=$(curl -sSf --max-time 30 "$css" | openssl dgst -sha384 -binary | openssl base64 -A) || {
-        echo "ERROR: Failed to fetch CSS from: ${css}" >&2
-        exit 1
-    }
-    echo "Hash: ${hash}"
+for i in "${!css_urls[@]}"; do
+    css="${css_urls[$i]}"
+    hash="${css_hashes[$i]}"
+    echo "Using pinned hash for: ${css}"
     css_links="${css_links}${css}|${hash}"$'\n'
 done
 css_links="${css_links%$'\n'}"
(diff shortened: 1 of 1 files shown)


yegor256:
Green, and pinning is the right principle. The reason I'm not merging it: the hashes were computed from whatever the CDN served at authoring time, and one of the seven URLs (the Zerocracy palette) is unversioned. The next publish of that file makes every generated page fail its integrity check and drop the stylesheet, with nothing in the build to warn about it. Worth deciding how that URL is handled before the pin lands. Assigning to @yegor256.

---

PR #874, opened 2026-09-15
#814: build the image from the lock file this repository keeps

### LINKED ISSUE #814: Docker image doesn't copy Gemfile.lock causing discrepancy in gem versions
## What happens

The `Dockerfile` copies the `Gemfile` into the image and not the `Gemfile.lock`:

```dockerfile
COPY Makefile /home
COPY makes /home/makes
COPY .stylelintrc.json /home
COPY Gemfile /home
COPY html-minifier-config.json /home
...
RUN make --directory=/home --no-silent install assets
```

and `makes/install.sh` runs a plain resolve:

```bash
bundle install
```

With no lock file present, Bundler resolves the whole graph from scratch against whatever is on rubygems.org at the moment the image is built. The `Gemfile.lock` that this repository commits, tests against and lets Renovate update has no effect on the artifact that users actually run.

## What the difference is

The constraints in the `Gemfile` are wide:

```ruby
gem 'factbase', '~>0.17'
gem 'fbe', '~>0.41'
gem 'judges', '~>0.59'
```

and the lock file pins:

```
factbase (0.19.16)
fbe (0.49.1)
judges (0.61.0)
```

So the image may ship any `factbase` from 0.17 up, any `fbe` from 0.41 up, any `judges` from 0.59 up. None of those combinations is the one the test suite ran against, except by coincidence on the day of the build.

These three gems are not incidental dependencies, they are the engine: `judges` runs the pipeline, `fbe` provides `Fbe.fb`, `Fbe.octo` and the bylaws, `factbase` provides the query language every judge and every stylesheet depends on. Behaviour differs between their patch releases. `Fbe.unmask_repos` in fbe 0.48.0, for instance, neither strips whitespace from a mask nor de-duplicates the resulting list, while a later one does both; a judge that works in CI can behave differently in the image for that reason alone.

The JavaScript half of the same script does pin, exactly:

```bash
npm --no-color install -g eslint@9.22.0
npm --no-color install -g uglify-js@3.19.3
npm --no-color install -g sass@1.77.2
npm --no-color install -g html-minifier@4.0.0
```

so the intent to have a reproducible image is already there; the Ruby half is the one that got left out.

## Why it matters

The failure mode is quiet. A user opens an issue saying the vitals page renders something odd, the maintainer cannot reproduce it because CI uses the lock, and the difference lives in an image that was built on a different day from a different resolve. `#628` already dealt with a version of this confusion, where CI tested a Docker image that was not the PR's code.

## Suggested fix

Copy the lock and install from it:

```dockerfile
COPY Gemfile /home
COPY Gemfile.lock /home
```

```bash
bundle config set --local deployment true
bundle install
```

`deployment true` makes Bundler fail when the `Gemfile` and the lock disagree, which turns a silent divergence into a build error at the moment somebody edits the `Gemfile` and forgets the lock.

If the lock is deliberately left out so the image picks up gem fixes without a release, that is a defensible choice, but then it belongs in a comment in the `Dockerfile`, because the presence of a committed `Gemfile.lock` says the opposite.

opusstudiohq-max:
Closes #814.

The Dockerfile copies `Gemfile` and not `Gemfile.lock`, and `makes/install.sh` runs a plain `bundle install`. With no lock inside the image, Bundler resolves the whole graph from scratch against whatever rubygems.org holds at build time.

So the lock file this repository commits, tests against, and lets Renovate update has **no effect on the artifact people actually run**. The Gemfile constraints are wide:

```ruby
gem 'factbase', '~>0.17'
gem 'fbe', '~>0.41'
gem 'judges', '~>0.59'
```

while the lock pins `0.19.16`, `0.49.1`, `0.61.0`. Two images built from the same commit on different days can ship different gems, and the versions the test suite proves are not the versions that run.

This is one more `COPY`, next to the one already there. The lock already carries `x86_64-linux` among its platforms, so nothing has to be re-resolved inside the image.

The docker build in `test.yml` exercises this directly, so CI here is the real check.

---


opusstudiohq-max:
Closing this as a duplicate of #832 by @gemshrine, which came first and covers the same defect.

My apologies for the noise. My duplicate screen compared PR **titles** against the issue, and your title describes the change rather than naming the number — the `Fixes #` reference is in the body, where I was not looking. That is my bug, not your title's. I have fixed the screen.

Nothing here is needed on top of #832.

PR #875, opened 2026-09-15
#820: replace the latest assessment in one transaction

### LINKED ISSUE #820: `latest-assessment` judge deletes the existing fact before a replacement is safely written
### Problem
`judges/latest-assessment/latest-assessment.rb` removes every existing `latest-assessment` fact with `delete!` and only then inserts the replacement.

### Why this is a bug
The judge performs a destructive operation before the write is known to have succeeded. Any exception while reading `latest.text`/`latest.when`, inserting the fact, or committing the factbase (for example a transient storage or validation failure) leaves the persistent factbase with no `latest-assessment` at all. Subsequent pages lose the latest assessment section even though the source `assessment` facts remain available. This also creates a window where readers can observe an empty result during a multi-step update.

### Expected behavior
Update an existing fact in place, or insert/validate the replacement before removing the old one, ideally within the factbase transaction semantics. Add a failure-path fixture proving the previous latest assessment survives an unsuccessful refresh.

opusstudiohq-max:
Closes #820.

The judge deleted every `latest-assessment` fact and *then* read `latest.text`, `latest.when` and inserted the replacement:

```ruby
Fbe.fb.query('(eq what "latest-assessment")').delete!

f = Fbe.fb.insert
f.what = 'latest-assessment'
...
```

The delete is already applied by the time the write is attempted, so anything raised in between leaves the factbase with **no** latest assessment at all — while the `assessment` facts it is derived from are still sitting right there. It also opens a window in which a reader sees nothing in the middle of an update.

`Factbase#txn` collects changes into a `LazyTaped` and applies them to the maps only *after* the block returns; an exception inside means the maps are never touched. Putting the delete and the insert in one block therefore makes the swap all-or-nothing.

### The failure-path test the issue asks for

`Factbase::Inv` forces the failure deterministically, raising on the `total` write — the last of the four — so the exception lands exactly in the window that used to be unsafe:

```ruby
fb = Factbase::Inv.new(Factbase.new) do |prop, _value|
  raise 'the storage refused this write' if prop == 'total'
end
```

Before this change the old fact is already gone by then and the test fails on a count of 0. After it, the stale assessment is still there, untouched.

The three existing packs keep passing, so the happy path is unchanged.

---


opusstudiohq-max:
Closing this as a duplicate of #846 by @twaldin, which came first and covers the same defect.

My apologies for the noise. My duplicate screen compared PR **titles** against the issue, and your title describes the change rather than naming the number — the `Fixes #` reference is in the body, where I was not looking. That is my bug, not your title's. I have fixed the screen.

Nothing here is needed on top of #846.

PR #876, opened 2026-09-16
#792: point the vitals page at its own URL with a canonical link

### LINKED ISSUE #792: The vitals page has no `canonical` link, though it knows its own URL
## What happens

The page carries no canonical link. `grep canonical xsl/*.xsl` finds nothing, while the page does
know its own address and already uses it:

```xml
<meta property="og:url">
  <xsl:attribute name="content">
    <xsl:value-of select="$url"/>
  </xsl:attribute>
</meta>
```

`url` is documented as "The URL where the static files will be visible", and the same value is
what the badge snippet at the bottom of the page hands out for copying.

GitHub Pages keeps serving `<owner>.github.io/<repo>/` after a custom domain is added, so the same
report is reachable at two addresses. Without a canonical link the two are indexed as separate
pages and split whatever ranking either would have had, and a search result can send a reader to
the address the project no longer uses.

## What should happen

The head should carry `<link rel="canonical" href="{$url}"/>` when `url` is given, pointing at the
one address the project published.

opusstudiohq-max:
GitHub Pages keeps serving `<owner>.github.io/<repo>/` after a custom domain is added, so the same report answers at two addresses and is indexed as two pages. The page already knows which one it is: `url` is documented as "The URL where the static files will be visible", `og:url` is built from it, and `entry.sh` falls back to `https://${GITHUB_REPOSITORY_OWNER}.github.io/${GITHUB_REPOSITORY#*/}` when the input is empty, so the value is there either way.

The head now carries `<link rel="canonical">` with that same `$url`, placed next to `og:url` since the two say the same thing to different readers.

It is guarded by `$url != ''`. The parameter defaults to an empty string in `action.yml`, and while `entry.sh` always fills it, the stylesheet is also driven directly by the tests, and `<link rel="canonical" href="">` would point a crawler at the directory it was served from. An empty `url` now yields no canonical link rather than a wrong one.

The test renders the whole page through Saxon with `url=https://example.com` — the fixture `generate_vitals_html` already uses — and reads the tag back with Nokogiri.

On `master` the assertion cannot pass: `git grep canonical origin/master -- xsl/` finds nothing, so `//link[@rel="canonical"]` is empty and `refute_nil` fails.

Verified on my fork first: `make` green, and the new test ran rather than skipping — `test_page_points_at_itself_with_a_canonical_link PASS (0.78s)`, 36 tests, 0 failures.

Closes #792

---


opusstudiohq-max:
Closing this until #792 is triaged. It carries no label, so the project has not accepted it as a defect yet, and a pull request opened ahead of triage asks for review work on a problem nobody has agreed to fix — and goes stale at a cost to the reviewers and reporters here, not only to me. Raised by @Thayorns on zerocracy/judges-action#2298, and it applies to my whole queue, not only to that repository. The branch stays on my fork; I will reopen this if and when #792 gets a label.

PR #877, opened 2026-09-16
#791: let an adless page preview with its own logo

### LINKED ISSUE #791: An `adless` page has no `og:image`, so a shared link previews as nothing
## What happens

The social preview image and its three size tags sit behind the `adless` flag:

```xml
<xsl:if test="$adless = 'false'">
  <meta property="og:image" content="https://www.zerocracy.com/og/vitals.png"/>
  <meta property="og:image:type" content="image/png"/>
  <meta property="og:image:width" content="1200"/>
  <meta property="og:image:height" content="630"/>
</xsl:if>
```

Dropping the Zerocracy picture from a white-labelled page is right. Leaving the page with no
picture at all is not: a link to an `adless` vitals page pasted into Slack, Telegram or a tweet
renders as a bare title and a sentence, while the same page with `adless: false` gets a card.

The project already hands the action its own image. `logo` is documented as "URL of the logo to
put on the HTML page" and is used for the favicon and the header, so the page knows what this
project looks like even when it may not show the Zerocracy one.

## What should happen

An `adless` page with a `logo` should use that logo as its `og:image`, so the link still previews
as something. An `adless` page without a `logo` keeps no image, as it does today.

opusstudiohq-max:
Dropping the Zerocracy picture from a white-labelled page is right; leaving the page with no picture at all is not. An `adless` vitals link pasted into Slack or a tweet rendered as a bare title while the same page with `adless: false` got a card.

The project already hands the action a picture of its own. `logo` is documented as "URL of the logo to put on the HTML page". The `xsl:if` becomes an `xsl:choose`: a non-`adless` page keeps the Zerocracy image and its three size tags exactly as before, and an `adless` page with a `logo` uses that logo as `og:image`. An `adless` page without a `logo` still gets no image, which is what the issue asks for.

The size and type tags stay behind the first branch. `og:image:width`, `og:image:height` and `og:image:type` describe the specific 1200x630 PNG at `zerocracy.com/og/vitals.png`; a project's own logo has dimensions this stylesheet does not know, and stating them wrongly is worse than omitting them — Open Graph consumers fetch the image when the tags are absent.

The test renders the page with `adless=true` through Saxon and reads `og:image` back. The existing fixture already passes `logo=x`, so the assertion is that the tag carries `x`.

On `master` the assertion cannot pass: the whole block sits behind `$adless = 'false'`, so an `adless` page emits no `og:image` element at all and `refute_nil` fails.

Verified on my fork first: `make` green, and the new test ran rather than skipping — `test_adless_page_previews_with_the_logo PASS (1.01s)`, 36 tests, 0 failures.

Closes #791

---


opusstudiohq-max:
Closing this until #791 is triaged. It carries no label, so the project has not accepted it as a defect yet, and a pull request opened ahead of triage asks for review work on a problem nobody has agreed to fix — and goes stale at a cost to the reviewers and reporters here, not only to me. Raised by @Thayorns on zerocracy/judges-action#2298, and it applies to my whole queue, not only to that repository. The branch stays on my fork; I will reopen this if and when #791 gets a label.

PR #878, opened 2026-09-16
#793: write what the run produced to the GitHub Actions run summary

### LINKED ISSUE #793: The action writes nothing to the GitHub Actions run summary
## What happens

The action writes nothing to `$GITHUB_STEP_SUMMARY`. `grep -r STEP_SUMMARY` over the repository
finds no use of it, so the summary page of a run is empty and everything the action has to say
lives in the log:

https://github.com/zerocracy/pages-action/blob/master/entry.sh

```bash
echo "The output directory is: ${INPUT_OUTPUT}"
...
echo "The HTML page is saved to: ${html}"
```

To find out whether a nightly run produced anything, or what it produced, a maintainer opens the
job log and reads it, or opens the published page. The run itself says nothing.

## What should happen

The run summary should carry what the run produced: the figures the page shows and a link to the
published page, when `url` is given. Written only when `GITHUB_STEP_SUMMARY` is set, so a local or
Docker run is unaffected.

opusstudiohq-max:
`grep -r STEP_SUMMARY` over the repository found nothing, so the summary page of a run was empty and everything the action had to say lived in the job log.

`entry.sh` now writes three things to `$GITHUB_STEP_SUMMARY` after the badge is generated: the factbase name, the figures, and links to what was published.

The figures are not recomputed. They are read back out of the page that was just written, from its `<meta name="description">` — which is exactly the sentence the page shows a reader: *"N average points per task, M total points earned, K contributors."* Recomputing them in shell would be a second implementation of `vitals.xsl` that could disagree with it. `html-minifier-config.json` sets only `collapseWhitespace`, `keepClosingSlash`, `removeComments` and `removeEmptyAttributes`, so the attribute quotes survive minification and the extraction is stable.

The links are built as `${url}/${name}-vitals.html` and `${url}/${name}-badge.svg`, the same shape the page's own copy-to-clipboard snippet uses at `xsl/vitals.xsl:387`. They are written only when `url` is non-empty, as the issue asks.

The whole block is guarded by `[ -n "${GITHUB_STEP_SUMMARY:-}" ]`, so a local or Docker run is untouched — the `:-` matters under `set -e -o pipefail`, since the variable is unset there rather than empty.

`entries/writes-the-run-summary.sh` runs the action with the variable pointed at a file and asserts the heading, the figures line and both links. Every other entry leaves the variable unset, so they also cover the quiet path.

Verified on my fork first: `make` green and the new entry ran — `👍🏻 writes-the-run-summary.sh passed`.

Closes #793

---


opusstudiohq-max:
Closing this until #793 is triaged. It carries no label, so the project has not accepted it as a defect yet, and a pull request opened ahead of triage asks for review work on a problem nobody has agreed to fix — and goes stale at a cost to the reviewers and reporters here, not only to me. Raised by @Thayorns on zerocracy/judges-action#2298, and it applies to my whole queue, not only to that repository. The branch stays on my fork; I will reopen this if and when #793 gets a label.

Issue #802, opened 2026-09-07
`copy-repo-names` judge permanently freezes repository metadata after first run


gemshrine:
## What happens

The `copy-repo-names` judge records repository details only once and never refreshes them. In `judges/copy-repo-names/copy-repo-names.rb:9-11`, every existing `repo-details` fact is collected into `known`, and its repository ID is removed from the IDs that will be fetched:

```ruby
known = Fbe.fb.query('(eq what "repo-details")').each.filter_map(&:repository).to_a
ids = Fbe.fb.query('(exists repository)').each.map(&:repository).uniq - known
return if ids.empty?
```

The inserted fact contains mutable values such as stars, forks, description, language, open issues, and `updated_at` (lines 20-31). Once a repository has one `repo-details` fact, later action runs skip the API call forever, so the repositories section publishes permanently stale statistics.

## Reproduction

1. Run the judge with repository `12345`; the API returns `stargazers_count: 1` and the factbase gets one `repo-details` fact with `stars = 1`.
2. Run it again after the repository has 2 stars (or a changed description/open-issue count).
3. The judge returns before fetching the repository because `12345` is in `known`; the factbase and generated page still show `stars = 1`.

The existing tests cover only insertion of a previously unseen repository and do not cover changed metadata on a subsequent run.

## Suggested fix

Refresh the existing `repo-details` fact (or replace it transactionally) on each successful API read, while preserving one current fact per repository so `repositories.xsl` does not render duplicates. Add a regression test that runs the judge twice with changed repository metadata and asserts that the second values are displayed.

gemshrine:
@morphqdd I can add a two-run fixture that changes stars and `open_issues_count`, then verify the rendered repositories section uses the refreshed values if useful.

0crat:
@gemshrine Thanks for reporting a new bug! You've earned +18 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+504](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #888, opened 2026-09-19
Refresh cached repository metadata

Voyagerroc-Lab:
Addresses #802. The judge now fetches metadata for all referenced repositories on each run, replaces any existing repo-details facts only after a successful API response, and leaves old facts intact for repositories whose requests fail. This prevents stale metadata and duplicate records.\n\nAdded a fixture with stale pre-existing metadata that asserts the updated values and exactly one fact remain.\n\nValidation: git diff --check passes. Ruby and Bundler are unavailable in this environment, so the repository's Ruby judge tests could not be run locally; the fixture is included for CI.


0crat:
@voyagerroc-lab It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`fix/refresh-cached-repository-metadata`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`887`". Your running score is [-6](https://zerocracy.github.io/judges-action/zerocracy-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

